PR #159, opened 2021-03-10
CI with GitHub actions

maximlt:
First attempt at using GitHub actions. I've decided to run the tests (lint, test, docs) with tox in the CI. This is different from what we were doing on GitLab but this seems more natural since locally we run the tests with tox.

The biggest difference so far with our previous setting is that there doesn't seem to be an integrated way to feed the tests and coverage reports to GitHub actions. I've looked at some popular repos and couldn't find anything interesting, it seems that in most cases the reports are just not handled in a specific way (one just has to go to the workflow logs to find some more details if need be). Some alternatives though:
* This action that adds a bot which can publish junit test reports: https://github.com/marketplace/actions/publish-unit-test-results
* The github docs show how to [upload test reports as an artifact](https://docs.github.com/en/actions/guides/building-and-testing-python#packaging-workflow-data-as-artifacts). I'm not sure this is of any interest though, besides that after that the report can be downloaded.
* Code coverage seems to be a little bit easier to handle (compared to test reports), with solutions like Codecov

To be done:
* [x] Port the Deploy workflow

Some ideas:
 * Since we use pre-commit locally, shouldn't we run it also on the CI? There's actually [an action](https://github.com/pre-commit/action) to simplify running pre-commit.
 * In GitLab we run the CI on Python 3.6 on linux. By expending the matrix we could run it on multiple versions of Python (not sure if it's required) and/or multiple OS (sounds more interesting, specially for windows).

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,9 +1,10 @@
-name: Publish to PyPI / GitHub
+name: Publish to PyPI
 
 on:
   push:
     tags:
       - "v*"
+  workflow_dispatch:
 
 jobs:
   build-n-publish:
@@ -20,7 +21,7 @@ jobs:
           python-version: "3.7"
 
       - name: Check that the current version isn't already on PyPi
-        run |
+        run: |
           if [ "$(./get_pypi_latest_version.sh)" != "$(python setup.py --version)" ]
           then
             echo "Current version is not on PyPI, proceed with bulding"
@@ -31,13 +32,13 @@ jobs:
 
       - name: Check long description is OK for PyPI with tox
         run: |
+          python -m pip install --upgrade pip
           python -m pip install tox
           tox -e pypi
 
       - name: Build a binary wheel and a source tarball
         run: |
-          python -m pip install --upgrade pip
-          pip install setuptools wheel
+          python -m pip install setuptools wheel
           python setup.py sdist bdist_wheel
       - name: Publish distribution to PyPI
         uses: pypa/gh-action-pypi-publish@release/v1
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,0 +1,46 @@
+name: Publish to PyPI / GitHub
+
+on:
+  push:
+    tags:
+      - "v*"
+
+jobs:
+  build-n-publish:
+    name: Build and publish to PyPI
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout source
+        uses: actions/checkout@v2
+
+      - name: Set up Python 3.7
+        uses: actions/setup-python@v2
+        with:
+          python-version: "3.7"
+
+      - name: Check that the current version isn't already on PyPi
+        run |
+          if [ "$(./get_pypi_latest_version.sh)" != "$(python setup.py --version)" ]
+          then
+            echo "Current version is not on PyPI, proceed with bulding"
+          else
+            echo "Current version is the latest version uploaded to PyPI"
+            exit 1
+          fi
+
+      - name: Check long description is OK for PyPI with tox
+        run: |
+          python -m pip install tox
+          tox -e pypi
+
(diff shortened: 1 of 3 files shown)


maximlt:
There is now a dedicated workflow to publish EODAG to PyPI. The condition is that there's a commit tagged with `v*`. It's not as strict as what we had on GitLab (`'$CI_COMMIT_TAG =~ /^v\d+\.\d+.*/'`) because the [docs say](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet) that this isn't a proper regular expression. We could replace this single filter by a set of filters instead of just one to match with our release tag patterns (v2.1.0, v2.0b1, v1.6.0rc1, ...).

The approach recommended by the action that uploads the distribution is to use an API token to authenticate to PyPI. @sbrunato if that's fine for you need to follow [this procedure](https://pypi.org/help/#apitoken), and add `PYPI_API_TOKEN` to the GitHub secrets (I have no access to the repo settings).

```yml
          user: __token__
          password: ${{ secrets.PYPI_API_TOKEN }}
```

We could add some more steps to automatically create a GitHub release and upload the distribution there. For example (from geopandas' publish workflow):
```yml
      - name: Create GitHub Release
        id: create_release
        uses: actions/create-release@v1
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }} # This token is provided by Actions, you do not need to create your own token
        with:
          tag_name: ${{ github.ref }}
          release_name: ${{ github.ref }}
          draft: false
          prerelease: false

      - name: Get Asset name
        run: |
          export PKG=$(ls dist/ | grep tar)
          set -- $PKG
          echo "name=$1" >> $GITHUB_ENV
      - name: Upload Release Asset (sdist) to GitHub
        id: upload-release-asset 
        uses: actions/upload-release-asset@v1
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
        with:
          upload_url: ${{ steps.create_release.outputs.upload_url }}
          asset_path: dist/${{ env.name }}
          asset_name: ${{ env.name }}
          asset_content_type: application/zip
```

Another approach would have been to run the workflow when we manually create a release on GitHub. I prefer the first approach though.

sbrunato (commented):
- `lint` may be set in a different workflow and run on all pushes without branch restriction, but not on pull requests.
- `tests` and `build-docs` should also be run on develop branch pushes
- `lint` may be replaced by https://github.com/pre-commit/action
- Why use `ubuntu` + `actions/setup-python` vs `python` image directly ?

maximlt:
The last updates:
* Linting now runs with pre-commit (black, isort, flake8, pycodestyle and more git hooks). It uses the pre-commit action which takes care of running pre-commit and caching the dependencies. Note that it runs with the option `--all-files`.
* The unit/integration tests run on both Python 3.6 and Python 3.9. Running the test suite on Windows was attempted but failed (#162). The community cache action is used to cache the PIP downloads.
* Building the docs also includes caching the PIP downloads.

About the usage of the cache action:

A cache must be given a cache key. Most of the examples online use a hash function provided by GitHub to create a hash from a file, e.g. a *requirements.txt*. This would work great with files including only pinned versions, which we don't have (eodag being a library, not an app). Instead I've used the hash of *setup.py* and the week number, this ensures that the cache is updated quite often (weekly at least) to pull the new versions of our dependencies. 

The lint / test / build_docs phase are now decoupled, while on GitLab test & build_docs depended upon the success of lint. We had a simpler linting job (just flake8) compared to now (flake8+black+isort...). In some cases it may be useful to merge a PR even if the lint job doesn't succeed.

The tox config file needed to be updated to run the command `mkdir` in the CI.

maximlt:
Note: the build_docs job failed for a while at the `sudo apt update` command. If that happens again, let's try to stick to ubuntu 18.04 instead of the latest which right now pulls ubuntu 20.04 and might be less stable.

sbrunato (approved):

PR #181, opened 2021-03-18
chore: tox fix for pypi

sbrunato:
`allowlist_externals = /bin/bash` in `tox.ini` was too restrictive and raised a warning (`/bin/bash` is used in local tests, but `/usr/bin/bash` in github actions).
See https://tox.readthedocs.io/en/latest/config.html#conf-allowlist_externals

Also added `pip install sphinx` to have rst2html.py utility available)


maximlt:
I had the same problem with `mkdir`.

Issue #133, opened 2020-04-30
handle search results pagination with a generator


sbrunato:
**[Original report](https://bitbucket.org/geostorm/eodag/issue/133) by me.**

----------------------------------------

In python API, use a generator / yield to automate the query of paginated search results

sbrunato:
Make `download_all()` method auto increment over pages to download all found pages at once.
- Add a parameter to `download_all()` to enable pages auto-iteration or not
- On search, save search params in `SearchResults` object the same way it is done at `EOProduct` level
- we may add a generator to `SearchResults` for auto-iteration, like a `.next_page()` method

sbrunato:
When using `download_all` on a paginated `SearchResult`, page iterator should be automatically used to fetch next pages products

PR #190, opened 2021-03-19
feat: add two new methods to search for products, search_iter_page and search_all

### LINKED ISSUE #191: Search for all the products matching some criteria with a given provider
We have had a lot of discussions on a private GitLab instance before we moved to GitHub about how to search/download all the products matching some criteria.

The problems we observed with the current API were:
* The method `search` returns a tuple of `(SearchResult, int)`. The first item contains some products, in most cases 20 if the user hasn't defined `items_per_page` in the search (20 being the default value) or less than 20 if less products match the criteria. The second item is the total number of products matching the criteria, it is obtained after sending a count request (or equivalent) to the provider. This number may in some cases just be an estimate, usually lower than the correct value (this is true for peps at the time of writing). So what happens is that the user actually gets the results from the first page only, but it's not so easy to understand (https://github.com/CS-SI/eodag/issues/149)
* We have the method `download_all` that does what it says, it downloads all the products contained in the `SearchResult` it takes as an argument. However, some users would expect that it just downloads all the products matching the criteria that were used to search them in the first place.
* There is no clear way or example about how to search (and download) for all the products matching some criteria. The current way would be to execute a `search` by incrementing the page value given as a parameter. The issue with this approach is that it would each time send a count request (whether it's a proper request or not depends on the underlying search plugin implementation, but in most cases it's currently the case), while this doesn't seem necessary. Or by increasing `items_per_page` to a very high value, which may be too high (providers have their own limit) and result in potentially an obscure error.

We have discussed different approaches:
1. Add some superpower to `download_all` so that it could download more than the products it would be given. The idea was to store in the `SearchResult` all the information required to run the same search, and as a consequence to run the search on the next page. The next products found would then have been downloaded. And so on until exhaustion of the results. It would have been used as follows: `products, _ = dag.search(**criteria); paths = dag.download_all(products, exhaust=True)`
2. Add a `search_all` method that would collect all the products matching some criteria.
  2.1. By adding to `SearchResult` the possibility to trigger a search for the next page (same idea as with `download_all`)
  2.2. By adding a `search_iter_page` method to `EODataAccessGateway` that would iterate over the pages of a search, and using that method internally to implement `search_all`.

The approach *2.2.* was chosen to be implemented (which doesn't necessarily mean the other approaches can't or won't be implemented if they prove to be required):
* Just adding more power to `download_all` felt like it was not sufficient. Since some providers do not return the exact count of products available in a `search`, it would have been impossible for these providers to know beforehand how many products  `download_all` would download. This feels a a little inappropriate.
* `search_all` happens to mirror `download_all` quite well, the usage of `products = dag.search_all(**criteria); paths = dag.download_all(products)` feels quite natural
* Exposing a method `search_iter_page` will help us to document and explain that (most) searches are paginated. This is not obvious to every user while this is important to understand how eodag works.

maximlt:
Fixes #191 

Two new methods for `EODataAccessGateWay`:
* `search_iter_page`: to iterate over the pages of a products search
* `search_all`: to get all the products matching some search criteria -> it uses `seach_iter_page` internally to iterate over all the pages that return products

`search_all` tries to be a little smarter than `search_iter_page` or  `search`, in the sense that it tries to minimize the number of requests made. It makes use of an internal parameter `max_items_per_page` configured per provider (see `providers.yml`). 8 providers have such a parameter configured, I have found these values either in their documentation (and checked it) or by trial and error. If this internal parameter isn't set, a default value of 50 is used. I chose it because it's the lowest one between all the providers (mundi) but it's still higher than the default items_per_page of 20 used by `search` originally. The user can also defined the value she wants to use dynamically (`search_all(items_per_page=...)`), which could come in handy if the value we have set doesn't work properly.

Note also that `search_iter_page` stops yielding products if the number of products found is less than the number of items requested. This spares one last request.

Next to that:
* The logic that prepared the search kwargs passed to `search_plugin.query` was extracted from `search` and put into a new private method `_prepare_search`. It's used in the three search methods.
* `test_core.py` was refactored to make it easier to add tests
* Lots of unit tests were added to `test_core.py`, for: `guess_by_product`, `_prepare_search`, `_do_search`, `search_iter_page` and `search_all`
* An end-to-end test was added for searching a product by ID (just a provider tested, sobloo)
* There is no a log warning emitted when `box` or `bbox` are used in any of the search methods. It recommends using `geom` instead

As discussed privately, there appears to be some search plugins or providers that do not support pagination. This is a problem for `search_iter_page` - and `search_all` as a consequence - since it means it would just iterate indefinitely. A solution to this problem was implemented, by comparing the first product of the next page with the first product of the previous one (their `properties["id"]` and `provider` values). The iteration stops if they are equal. There is a test for this approach, which I have also tested manually with `astraea_eod` whose API only implements a *limit* and not a *page*.

I plan to add a proper documentation to these two new methods in a following PR, since the tutorials need to be refactored a little.

### DIFF
--- a/eodag/api/core.py
+++ b/eodag/api/core.py
@@ -556,6 +556,79 @@ class EODataAccessGateway(object):
             search_plugin, count=True, raise_errors=raise_errors, **search_kwargs
         )
 
+    def search_iter_page(
+        self,
+        items_per_page=DEFAULT_ITEMS_PER_PAGE,
+        start=None,
+        end=None,
+        geom=None,
+        locations=None,
+        **kwargs
+    ):
+        """Iterate over the pages of a product search.
+
+        :param items_per_page: The number of results that must appear in one single
+                               page (default: 20)
+        :type items_per_page: int
+        :param start: Start sensing UTC time in iso format
+        :type start: str
+        :param end: End sensing UTC time in iso format
+        :type end: str
+        :param geom: Search area that can be defined in different ways:
+
+                    * with a Shapely geometry object:
+                      ``class:`shapely.geometry.base.BaseGeometry```
+                    * with a bounding box (dict with keys: "lonmin", "latmin", "lonmax", "latmax"):
+                      ``dict.fromkeys(["lonmin", "latmin", "lonmax", "latmax"])``
+                    * with a bounding box as list of float:
+                      ``[lonmin, latmin, lonmax, latmax]``
+                    * with a WKT str
+
+        :type geom: Union[str, dict, shapely.geometry.base.BaseGeometry]
+        :param locations: Location filtering by name using locations configuration
+                          ``{"<location_name>"="<attr_regex>"}``. For example, ``{"country"="PA."}`` will use
+                          the geometry of the features having the property ISO3 starting with
+                          'PA' such as Panama and Pakistan in the shapefile configured with
+                          name=country and attr=ISO3
+        :type locations: dict
+        :param dict kwargs: some other criteria that will be used to do the search,
+                            using paramaters compatibles with the provider
+        :returns: An iterator that yields page per page a collection of EO products
+                  matching the criteria
+        :rtype: Iterator[:class:`~eodag.api.search_result.SearchResult`]
+
+        .. versionadded::
+            2.2.0
+        """
+        search_kwargs = self._prepare_search(
+            start=start, end=end, geom=geom, locations=locations, **kwargs
+        )
+        search_plugin = search_kwargs.pop("search_plugin")
+        search_kwargs.update(
+            page=1,
+            items_per_page=items_per_page,
+        )
+        while True:
+            logger.debug("Iterate over pages: search page %s", search_kwargs["page"])
+            products, _ = self._do_search(
+                search_plugin, count=False, raise_errors=True, **search_kwargs
+            )
+            if len(products) > 0:
+                yield products
+                # Prevent a last search if the current one returned less than the
+                # maximum number of items asked for.
+                if len(products) < items_per_page:
+                    last_page_with_products = search_kwargs["page"]
+                    break
+            else:
+                last_page_with_products = search_kwargs["page"] - 1
+                break
+            search_kwargs["page"] += 1
+        logger.debug(
+            "Iterate over pages: last products found on page %s",
+            last_page_with_products,
+        )
+
     def _search_by_id(self, uid, provider=None):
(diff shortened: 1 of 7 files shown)


sbrunato:
For `astraea_eod`, STAC API pagination only needs `item["links"][]["rel"] == "next"`. `page` parameter is only optional, see https://github.com/radiantearth/stac-api-spec/tree/master/item-search#paging.

For STAC providers, we should this `next` entry by default instead of `page`+`limit`.

This could be done by adding a new parameter in `providers.yml` / `stac_provider.yml`:
```yml
  pagination:
    next_page_url_key_path: '$.links[?(@.rel="next")].href'
```
And use `jsonpath_ng.ext.parse` instead of `jsonpath_ng.parse`, see https://github.com/h2non/jsonpath-ng/issues/8

maximlt:
@sbrunato I have implemented a mechanism to retrieve the next page when a provider doesn't implement "page".

In `search_iter_page` a search plugin is introspected to check whether it has a `pagination.next_page_url_key_path` key. If it does, it is assumed that it doesn't implement "page", and a kwarg `set_next_page_url` is passed down to `.query`. This is handled within `QueryStringSearch.do_search` where the response (JSON only right now) is parsed to get the "next" URL. It replaces the configuration key `pagination.next_page_url_tpl`. So the next call to `.query` will use that URL and will update it to get the next page.

I have added comments since this logic is quite special. Maybe we'll come up with a better way in the future.

maximlt:
@sbrunato I've made the changes as we agreed:
* `search_iter_page` doesn't propagate a boolean to trigger to "next" search through the plugin infrastructure down to `do_search`. Instead `do_search` takes care of setting an attribute `next_page_url` if the pagination config has the key `next_page_url_key_path`.
* `search_iter_page` uses that URL after the first iteration if available. It uses it by replacing `search_plugin.config.pagination["next_page_url_tpl"]`
* `search_plugin.config.pagination["next_page_url_tpl"]` and `search_plugin.next_page_url` are both reset after each iteration. The generator holds the `next_page_url` value in its internal state. This has the advantage that the search plugin is in a clean state after `next(dag.search_iter_page(**search_criteria))`. I thought it was quite important that `search_plugin.next_page_url` was reset every time, I didn't want a URL from a previous search to mess up with a new search (different criteria).

sbrunato (approved):

sbrunato:
updates #133

PR #192, opened 2021-03-20
Add Windows support

### LINKED ISSUE #162: Windows support
I've had the opportunity to run the test suite on Windows while porting the CI to GitHub actions. The logs of a workflow run on Windows with Python 3.6 can be found [here](https://github.com/CS-SI/eodag/runs/2113132127?check_suite_focus=true). A few tests are failing, mostly due to how the paths are handled within the tests themselves. A deeper analysis is required though, there may be some edge cases.

For an unknown reason, the tests [don't seem to run](https://github.com/CS-SI/eodag/runs/2113132205?check_suite_focus=true) on Windows with Python 3.9.

It'd be nice to support Windows, particularly if there's one day a QGIS plugin that uses EODAG internally (I think that most QGIS users use Windows).

maximlt:
Fixes #162 

Small updates required to add Windows support:
* `tempfile.NamedTemporaryFile` doesn't work exactly the same on Windows: https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile
* One test expected a Unix "/tmp" directory
* The template path for the locations shapefile expected a `/` path separator

The tests now also run on Windows.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -28,8 +28,8 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        python-version: [3.6, 3.7, 3.8, 3.9]
-        os: [windows-latest]
+        python-version: [3.6, 3.9]
+        os: [ubuntu-latest, windows-latest]
     steps:
     - name: Checkout the repo
       uses: actions/checkout@v2
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -28,8 +28,8 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        python-version: [3.6, 3.9]
-        os: [ubuntu-latest, windows-latest]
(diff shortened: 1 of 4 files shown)


maximlt:
The test were green on Windows/Python3.9 while they were actually not running at all. The tests discovery seemed broken.

I've run the Windows test suite on each Python version from 3.6 to 3.9. 3.6/3.7 run fine while 3.8/3.9 had the discovery problem.
I have found a [workaround on an SO answer](https://stackoverflow.com/questions/58556183/nose-unittest-discovery-broken-on-python-3-8): add `--traverse-namespace` to the `nosetests` command options.
There's an issue opened on nose (https://github.com/nose-devs/nose/issues/1106) that seems to spot the problem, the PR submitted was never merged (nose seems inactive).

I have the feeling that at some point we will have to update EODAG's test suite, more problems like this one are ahead of us this nose doesn't get any update unfortunately.

Note that the tests on Windows seem to take a little longer to run than those on Ubuntu (~4 min vs ~3 min).

sbrunato (approved):

PR #218, opened 2021-04-01
feat: SAFE build for STAC AWS providers

sbrunato:
fixes #171

PR #228, opened 2021-04-12
feat: sentinel non-safe product types

### LINKED ISSUE #216: Sentinel SAFE / non-SAFE product types distinction
All providers do not provide Sentinel products in SAFE format, which means for example that searching `S2_MSI_L1C` products on `peps` or `astraea_eod` can return products packaged differently, and differently formatted ids.

We should add more product types to be able to distinguish:
- Sentinel products in SAFE / jp2 format (S2_MSI_L1C peps)
- Sentinel products in non-SAFE / jp2 format (S2_MSI_L1C astraea_eod)
- Sentinel products in cog format (sentinel-s2-l2a-cogs earth_search)

See https://github.com/CS-SI/eodag/issues/136#issuecomment-808082569, #171

sbrunato:
Add new Sentinel non-SAFE product types, fixes #216

And updates AWS download SAFE formatting, please note that :
- to keep the ability to search by id on AWS STAC providers (_astraea_eod_ and _earth_search_), product id will not be SAFE formatted but will use the formatting used by the provider (see also #218). 
I.e. `id="S2B_31TCH_20210318_0_L1C"` for _earth_search_  and `id="S2B_OPER_MSI_L1C_TL_EPAE_20201201T115506_A019525_T30TYP_N02.09"` for _astraea_eod_
- If the product title (used as download destination path) can be SAFE formatted from search results, it will be. 
For example with _earth_search_ we can have `product.properties["id"]="S2B_31TCH_20210318_0_L1C"` and `product.properties["title"]="S2B_MSIL1C_20210318T103649_N0209_R008_T31TCH_20210318T143809"`. Destination path will be `"{title}/{title}.SAFE"`
- If the product title needs and extra request to be SAFE formatted, as for _astraea_eod_, this will not be done during search, but during download. 
The product destination will be in this case: `"{old_title}/{safe_title}.SAFE"` with `old_title=product.properties["title"]="S2B_OPER_MSI_L1C_TL_EPAE_20201201T115506_A019525_T30TYP_N02.09"` and `safe_title="S2B_MSIL1C_20201201T105419_N0209_R051_T30TYP_20201201T115506"`


maximlt:
Just a small note:

The new product types (JP2, COG) differ from the SAFE-formatted one in `product_types.yml` by their abstract (a mention was added at the end). Would it make sense to add a new `format` field (or other) in the catalog? Or are the existing fields already following some standard?

I'm thinking about the fact that the ability to search through this catalog with whoosh is a nice feature of eodag. `guess_product_type` could at some point be a little more exposed and support more query parameters.

sbrunato:
> Would it make sense to add a new format field (or other) in the catalog?

Yes it would. But we may think of what information we should put in that field for non-SAFE and non-Sentinel product types

> are the existing fields already following some standard?

We try to use as much as possible parameters from _OGC EO extension_ but some were not available (`missionStartDate` for example)

Issue #234, opened 2021-04-13
Error loading user config files with settings of providers loaded via plugins


ahuarte47:
**Install sentinelsat plugin**
```
python -m pip install eodag
python -m pip install eodag-sentinelsat
```

Append your own "scihub" provider settings to an user configuration file...

![image](https://user-images.githubusercontent.com/5576272/114627081-f564c380-9cb4-11eb-990b-2aa901040207.png)

**Code To Reproduce**
CLI command:
```sh
eodag --verbose search --conf "eodag-user_conf.yml" ....
```
**Output**
![image](https://user-images.githubusercontent.com/5576272/114627430-86d43580-9cb5-11eb-80af-5c9b5a398734.png)

**Environment:**
 - Python version: 3.7.4
 - EODAG version: 2.2.0

sbrunato:
Hello, many thanks for this PR!

PR #235, opened 2021-04-13
[Bugfix #234]: Error loading user config files with settings of providers loaded via plugins

ahuarte47:
This PR is a proposal to fix the issue https://github.com/CS-SI/eodag/issues/234.

It would nice we can add to user configuration files our own settings of providers loaded via plugins. 

This minor change allows we configure all providers in the same way, not taking care if they are implemented in the code base or via plugins.


sbrunato:
Hello, many thanks for this PR!

sbrunato (approved):

ahuarte47:
Thank you very much @sbrunato  !!!

PR #276, opened 2021-04-30
Progress callback unify

### LINKED ISSUE #237: Inconsistent progress bar between download and download_all
**In a notebook**, I've observed that `EODataAccessGateway.download_all` was automatically selecting the right progress bar while `EODataAccessGateway.download` wasn't.

It'd be nice if `EODataAccessGateway.download` could also select automatically the right progress bar. We could then completely remove all the `from eodag.utils import NotebookProgressCallback; dag.download(product, progress_callback=NotebookProgressCallback())` from the notebooks :+1:

sbrunato:
fixes #237 

# Changes:

- `eodag.utils.ProgressCallback` now inherits from [`tqdm.auto.tqdm`](https://github.com/tqdm/tqdm/#documentation), and accepts the same arguments on instantiation.
- Deprecated `eodag.utils.get_progress_callback` and `eodag.utils.NotebookProgressCallback` (use `eodag.utils.ProgressCallback` instead)
- `Download._prepare_download()` and `Download._finalize()` now now have an optional `progress_callback` parameter.
- Updated download methods call graph: 
  - `Download.download_all()` now calls `product.download()` instead of  `plugin.download()`
  - Default `progress_callback` gets instantiated and closed in `product.download()`, to facilitate its handling at plugin level. 
image

- Using `EODataAccessGateway.download_all(products)`:
  - Creates a bar to track downloaded products count progress
  - Creates a bar per product to track individual download and extraction (same bar for download & extract)
image
- Using `EODataAccessGateway.download_all(products, progress_callbak=ProgressCallback())`:
  - Uses given progress_callbak to track downloaded products count progress
  - Uses a copy of given progress_callbak to track individual download and extraction (same bar for all products download & extract). Progress bar stays opened until user closes it.
image
- Progress bar can be disabled using `ProgressCallback(disable=True)`
- Updated `eodag.setup_logging` to permit global progress bars disable:
```py
def setup_logging(verbose, no_progress_bar=False):
    """Define logging level

    :param verbose: Accepted values:

                    * 0: no logging with muted progress bars
                    * 1: no logging but still displays progress bars
                    * 2: INFO level
                    * 3: DEBUG level
    :type verbose: int

    :param no_progress_bar: (optional) disable progress bars (default=False)
    :type no_progress_bar: bool
    """
```


eodag/utils/logging.py
```diff
@@ -18,18 +18,27 @@
...
+                    * 0: no logging with mutted progress bars
+                    * 1: no logging but still displays progress bars
+                    * 2: INFO level
                     * 3: DEBUG level
     :type verbose: int
+
+    :param verbose: (optional) disable progress bars (default=False)
+    :type verbose: bool
     """
-    if verbose == 0:
```
maximlt, line 32: I'm not sure I get it, what's the difference between setup_logging(0) and setup_logging(1)?

eodag/utils/__init__.py
```diff
@@ -399,23 +443,11 @@ def __exit__(self, *exc):
...
-                unit_scale=self.unit_scale,
-                desc=self.desc,
-                position=self.position,
-                dynamic_ncols=True,
-            )
-        self.pb.update(current_size)
+    pass
 
 
 def get_progress_callback():
```
maximlt, line 453: When used in an Ipython environment, it will still return `NotebookProgressCallback` and emit a deprecated warning.

maximlt (commented):
Interesting approach @sbrunato  (it actually makes the `progress_callback` parameter pretty useless). And yes I think it fixes #237, but there's still in the code base this mixed usage of `bar = ProgressCalback()` and `bar = get_progress_callback()`. I guess with theses changes `get_progess_callback()` is of no use now. A couple more things:
* The docs will need to be updated at some point (NotebookProgressCallback is in every download commands)
* The CLI assumes `verbosity=0`, which if I get it right means that there is no progress bar. I guess there should be a progress bar by default.

eodag/utils/__init__.py
sbrunato: thanks for the feedback @maximlt 
> there's still in the code base this mixed usage of bar = ProgressCalback() and bar = get_progress_callback()

> When used in an Ipython environment, it will still return NotebookProgressCallback and emit a deprecated warning.

- [ ] Deprecate `get_progress_callback` and make it always return `ProgressCalback`

> The docs will need to be updated

- [ ] update the doc and remove custom progress callback usage or make it use `ProgressCalback`

> what's the difference between setup_logging(0) and setup_logging(1)?

Same logging level, with progress-bar enabled when using  `setup_logging(1)`, and disabled when using `setup_logging(0)`

PR #317, opened 2021-07-09
feat: new notebook extra dependency

sbrunato:
Adds new `notebook` extra dependency which will install  `ipywidgets` through `tqdm[notebook]`:
```sh
pip install eodag[notebook]
```

Please note that `pip >= 20.2` will be needed to install this extra as `tqdm` is already in `install_requires`, see https://github.com/pypa/pip/issues/988

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -74,6 +74,9 @@ setup(
             "flake8",
             "pre-commit",
         ],
+        "notebook": [
+            "tqdm[notebook]",
+        ],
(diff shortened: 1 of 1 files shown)


maximlt:
@sbrunato about the conda-forge distribution I think `ipywidgets` could manually be added in the recipe, since I assume conda users are more likely to use eodag in a notebook.

PR #344, opened 2021-09-03
Fix get_quicklook method for onda provider

drnextgis:
This PR aims to solve the issue of getting overviews for ONDA provider:
```
TypeError                                 Traceback (most recent call last)
<ipython-input-7-3c773c92be6c> in <module>
      1 workspace = "eodag_workspace_overview"
      2 quicklooks_dir = os.path.join(workspace, "quicklooks")
----> 3 product.get_quicklook(base_dir=quicklooks_dir)

~/eodag/eodag/api/product/_product.py in get_quicklook(self, filename, base_dir, progress_callback)
    399             ):
    400                 with open(quicklook_file, "wb") as fd:
--> 401                     fd.write(self.properties["quicklook"])
    402                 return quicklook_file
    403 

TypeError: a bytes-like object is required, not 'str'
```

The quicklook in this case is a [base64](https://gist.github.com/drnextgis/e494a1150fffda1189c80362312b9275) string which should be decoded.

### DIFF
--- a/eodag/api/product/_product.py
+++ b/eodag/api/product/_product.py
@@ -15,6 +15,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import base64
 import logging
 import os
 import re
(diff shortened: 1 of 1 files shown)


sbrunato:
@drnextgis many thanks for this PR!

drnextgis:
Is there anything preventing this PR from being merged?

sbrunato:
@drnextgis no problem, it is done. Thanks again!

PR #357, opened 2021-10-29
Download callback

### LINKED ISSUE #121: download_all callback
**[Original report](https://bitbucket.org/geostorm/eodag/issue/121) by me.**

----------------------------------------

When using download_all, provide the availability to launch a callback when each individual download in finished

CGuichard:
Closes #121.

- For `download_all`: callback is done inside the `download_all` method of `eodag.plugins.download.base.Download`, because its children  `download_all` methods delegates to `super()`. Every class inheriting `Download` will get this callback behavior when calling `super().download_all(...)`.
- For `download`: the callback is done after the download is finished, inside the `download` method of `EOProduct`. i wanted to use it too inside of the `Download` class but its children don't call its `download` method. I could add the callback inside every children `download` method, but some of them have a lot or `return`, it is not really efficient.

As of now, callback is inside of `Download` for `download_all`, and in `EOProduct` for `download`.
While it is working, it's not clean. These are the changes I see to unify the logic behind the callback location :

- Just like for `download_all`, by making the class inheriting `Download` call their parent `download` I could add this behavior

OR

- Make the callback inside of `download` and `download_all` of `EODataAccessGateway`. Here, the callback logic is high level and is not present in any of the `eodag.plugins.apis.base.Api`  and `eodag.plugins.download.base.Download` classes.

### DIFF
--- a/docs/api_reference/utils.rst
+++ b/docs/api_reference/utils.rst
@@ -10,9 +10,10 @@ Logging
 .. automodule:: eodag.utils.logging
    :members:
 
-Progress callback
+Callbacks
 -----------------
 
+.. autofunction:: eodag.utils.AllDownloadedCallback
 .. autofunction:: eodag.utils.ProgressCallback
 
 Notebook
--- a/docs/api_reference/utils.rst
+++ b/docs/api_reference/utils.rst
@@ -13,6 +13,7 @@ Logging
 Callbacks
 -----------------
 
+.. autofunction:: eodag.utils.DownloadedCallback
 .. autofunction:: eodag.utils.AllDownloadedCallback
 .. autofunction:: eodag.utils.ProgressCallback
 
--- a/eodag/utils/__init__.py
+++ b/eodag/utils/__init__.py
@@ -399,6 +399,18 @@ def get_timestamp(date_time):
     return dt.timestamp()
 
 
+class AllDownloadedCallback:
+    """Example class for :meth:`~eodag.api.core.EODataAccessGateway.download_all` callback"""
+
+    def __call__(self, search_result):
+        """Callback
+
+        :param search_result: A collection of EO products resulting from a search
+        :type search_result: :class:`~eodag.api.search_result.SearchResult`
+        """
+        logger.debug("Download finished for the search result %s", search_result)
+
+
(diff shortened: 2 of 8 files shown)


sbrunato (changes requested):
I suggest to change some points:
- rename `download_callback` parameter to `downloaded_callback`
- rename `DownloadProductCallback` class to `DownloadedCallback`
- rename `DownloadSearchResultCallback` class to `AllDownloadedCallback`

And for the location of the callback execution, having a Iook at the download methods call graph in the [documentation](https://eodag.readthedocs.io/en/latest/plugins_reference/download.html#download-plugins)(scroll down), I would suggest:
- `EODataAccessGateway` for `download_all()`
- `EOProduct` for `download()`

eodag/utils/__init__.py
```diff
@@ -399,6 +399,30 @@ def get_timestamp(date_time):
     return dt.timestamp()
 
 
+class DownloadedCallback:
+    """Example class for download() callback"""
```
sbrunato, line 403:
```suggestion
    """Example class for :meth:`~eodag.api.core.EODataAccessGateway.download` callback"""
```

eodag/utils/__init__.py
```diff
@@ -399,6 +399,30 @@ def get_timestamp(date_time):
...
+        """Callback
+
+        :param product: The EO product downloaded
+        :type product: :class:`~eodag.api.product._product.EOProduct`
+        """
+        logger.debug("Download finished for the product %s", product)
+
+
+class AllDownloadedCallback:
+    """Example class for download_all() callback"""
```
sbrunato, line 415:
```suggestion
    """Example class for :meth:`~eodag.api.core.EODataAccessGateway.download_all` callback"""
```

docs/api_reference/utils.rst
```diff
@@ -13,6 +13,8 @@ Logging
 Progress callback
```
sbrunato, line 13:
```suggestion
Callbacks
```

sbrunato (changes requested):
We can add tests for callbacks usage with `download()` and `download_all()`

sbrunato (approved):

PR #359, opened 2021-11-03
Attach the crunchers directly to SearchResult

### LINKED ISSUE #206: Attach the crunchers directly to `SearchResult`?
As of now there are two ways to apply the crunchers (out of the CLI):
1. With `SearchResult.crunch(cruncher, **search_params)`. One example can be found in the advanced tutorial:
```python
from eodag.plugins.crunch.filter_overlap import FilterOverlap

filtered_products = products.crunch(FilterOverlap({"minimum_overlap": 100}), geometry=extent)
```
2. With `EODataAccessGateway.crunch(results, **kwargs)`. There's no example of this usage in the tutorials. It's however used internally to implement the `eodag search` command. Its advantage is that it allows to run multiple crunchers sequentially with a single method call.
```python
dag = EODataAccessGateway()
criteria = dict(productType="xxx", start="xxx", end="xxx", geom="xxx")
products, _ = dag.search(**criteria)
crunchers = {
    "FilterDate": {"start": "2020-08-10", "end": "2020-08-12"},
    "FilterOverlap": {"intersects": True}
}
filtered_prods = dag.crunch(products, search_criteria=dict(geometry=criteria["geom"]), **crunchers)
```

The second way of using the crunchers seems overly complicated, I doubt there is much usage of this except the internal case already mentioned.

The first way is a little easier, even if it has some problems:
* It requires to import a cruncher to apply it
* It requires to create an instance of this cruncher, whose constructor requires a dictionary, whose expected content is written in the class docstring. So using this in an IDE doesn't show what key/value items the dictionary should have (with ipython I do `FilterOverlap?` to get that info).
* The `.crunch` method of a `SearchResult` sometimes requires that some additional kwargs are passed. For instance `FilterOverlap` requires `geometry=whatever`. This is included in the `.proceed` method of `FilterOverlap` and can only be accessed through the API documentation.

As shown above, using the crunchers we provide is probably quite complicated. This is unfortunate since this kind of filter can be really handy.  For example (there are certainly many more examples):
* Google Earth Engine proposes similar filters with (`filterDate`, `filterBounds`, `filterMetadata` or `sort`) that can be applied to an `ee.ImageCollection`.
* Planet Labs proposes in its Python client a [series of filters](https://planetlabs.github.io/planet-client-python/api/reference.html#filters) including among others  `and_filter`, `or_filter`, `date_range` and `geom_filter`.

However, as I understand it, these two examples apply the filters server-side, contrary to eodag that applies them client-side (after the products are collected). This has pros:
* We can offer as many filters as we want, without depending on what the providers support or not
* An advanced user can develop her own filter since a cruncher can be added as a plugin, similarly to a Search or Download plugin

And cons:
* The request made to a provider is not as narrow as it could be if a user wants to apply a filter that the requested provider supports (I'm not yet informed enough about what the providers support or not to give a more precise example here, I just assume some support more than 'start', 'end', 'geometry'). However, eodag will propagate additional search criteria provider by an user (`cloudCover=10` for instance), so an advanced user could make a search with some specific provider's filters if she knows how to configure them.

In order to make the crunchers easier to use, we could add them individually as methods of `SearchResult`. This would allow us to give them some proper signature. It would be interesting that if instead of returning a bare list they would return a `SearchResult`:
```python
prods = dag.search_all(**criteria)
filtered_prods = prods.filter_date(start="xxx", end="xxx")
filtered_prods = filtered_prods.filter_bounds(predicate="within", geometry="POLYGON ...")
```
Note that the filters could be chained in the example above, since each filter is expected to return a `SearchResult`

An example implementation would be:
```python
from eodag.plugins.crunch.filter_date import FilterDate as _FilterDate
class SearchResult
    def crunch(...):
        ...
    def filter_date(start=None, end=None):
        filtered_prods = self.crunch(_FilterDate(dict(start=start, end=end)))
        return SearchResult(filtered_prods)
```

CGuichard:
Closes #206.

### DIFF
--- a/eodag/api/search_result.py
+++ b/eodag/api/search_result.py
@@ -18,6 +18,11 @@
 from collections import UserList
 
 from eodag.api.product import EOProduct
+from eodag.plugins.crunch.filter_date import FilterDate
+from eodag.plugins.crunch.filter_latest_intersect import FilterLatestIntersect
+from eodag.plugins.crunch.filter_latest_tpl_name import FilterLatestByName
+from eodag.plugins.crunch.filter_overlap import FilterOverlap
+from eodag.plugins.crunch.filter_property import FilterProperty
 
 
 class SearchResult(UserList):
@@ -42,6 +47,58 @@ class SearchResult(UserList):
         """
         crunched_results = cruncher.proceed(self, **search_params)
         return SearchResult(crunched_results)
+
+    def filter_date(self, start=None, end=None):
+        """
+        Apply :class:`~eodag.plugins.crunch.filter_date.FilterDate` crunch,
+        check its documentation to know more.
+        """
+        return self.crunch(FilterDate(dict(start=start, end=end)))
+
+    def filter_latest_intersect(self, geometry):
+        """
+        Apply :class:`~eodag.plugins.crunch.filter_latest_intersect.FilterLatestIntersect` crunch,
+        check its documentation to know more.
+        """
+        return self.crunch(FilterLatestIntersect({}), geometry=geometry)
+
+    def filter_latest_by_name(self, name_pattern):
+        """
+        Apply :class:`~eodag.plugins.crunch.filter_latest_tpl_name.FilterLatestByName` crunch,
+        check its documentation to know more.
+        """
+        return self.crunch(FilterLatestByName(dict(name_pattern=name_pattern)))
+
+    def filter_overlap(
+        self,
+        geometry,
+        minimum_overlap=0,
+        contains=False,
+        intersects=False,
+        within=False,
+    ):
+        """
+        Apply :class:`~eodag.plugins.crunch.filter_overlap.FilterOverlap` crunch,
+        check its documentation to know more.
+        """
+        return self.crunch(
+            FilterOverlap(
+                dict(
+                    minimum_overlap=minimum_overlap,
+                    contains=contains,
+                    intersects=intersects,
+                    within=within,
+                )
+            ),
+            geometry=geometry,
+        )
+
+    def filter_property(self, operator="eq", **search_property):
+        """
+        Apply :class:`~eodag.plugins.crunch.filter_property.FilterProperty` crunch,
+        check its documentation to know more.
+        """
+        return self.crunch(FilterProperty(dict(operator=operator, **search_property)))
 
     @staticmethod
     def from_geojson(feature_collection):
--- a/docs/api_reference/searchresult.rst
+++ b/docs/api_reference/searchresult.rst
@@ -17,6 +17,11 @@ Crunch
 .. autosummary::
 
    SearchResult.crunch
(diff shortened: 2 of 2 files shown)


eodag/api/search_result.py
```diff
@@ -42,6 +47,43 @@ def crunch(self, cruncher, **search_params):
...
+        """Apply FilterLatestByName crunch, check its documentation to know more"""
+        return self.crunch(FilterLatestByName(dict(name_pattern=name_pattern)))
+
+    def filter_overlap(
+        self,
+        geometry,
+        minimum_overlap=None,
+        contains=None,
+        intersects=None,
+        within=None,
```
sbrunato, line 68:
```suggestion
        minimum_overlap=0,
        contains=False,
        intersects=False,
        within=False,
```

sbrunato (changes requested):

eodag/api/search_result.py
```diff
@@ -42,6 +47,43 @@ def crunch(self, cruncher, **search_params):
...
+                    minimum_overlap=minimum_overlap,
+                    contains=contains,
+                    intersects=intersects,
+                    within=within,
+                )
+            ),
+            geometry=geometry,
+        )
+
+    def filter_property(self, operator=None, **search_property):
```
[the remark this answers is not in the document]
CGuichard: This method `filter_property` delegates to `FilterProperty`. The default values are managed in the class, by using `None` the default values inside `FilterProperty` are used. 
If we add "eq" here there will be redundancy with the implementation of `FilterProperty`. It will help users to know what is the value used by `filter_property` without going to check the doc of `FilterProperty`, but if we change the default value inside `FilterProperty` we will need to also change it here, across multiple files, for one value changed.

eodag/api/search_result.py
CGuichard: My reasoning here is the same as for `filter_property`

eodag/api/search_result.py
```diff
@@ -43,6 +48,43 @@ def crunch(self, cruncher, **search_params):
         crunched_results = cruncher.proceed(self, **search_params)
         return SearchResult(crunched_results)
 
+    def filter_date(self, start=None, end=None):
+        """Apply FilterDate crunch, check its documentation to know more"""
+        return self.crunch(FilterDate(dict(start=start, end=end)))
+
+    def filter_latest_intersect(self, geometry):
+        """Apply FilterLatestIntersect crunch, check its documentation to know more"""
```
sbrunato, line 56:
```suggestion
        """Apply :class:`~eodag.plugins.crunch.filter_latest_intersect.FilterLatestIntersect` crunch, check its documentation to know more"""
```

sbrunato (changes requested):

sbrunato (approved):

PR #458, opened 2022-05-16
Online filter

### LINKED ISSUE #215: Avoid trying to order an OFFLINE product
From @remi-braun in https://github.com/CS-SI/eodag/issues/214

> Maybe add the LTA order as a parameter, as it is not needed in every usecase (sometimes, sb just want to download everything he can without triggering any LTA retrieval)

LTA: Long Term Archive

The `HTTPDownload` plugin (used by creodias, onda, theia, peps, sobloo) can automatically try to order a product if its status is *OFFLINE*, and can download it afterwards.

The feature request is to add a way to avoid trying to order an *OFFLINE* product.

This could be done in at least two complementary ways:
1. Document the way to filter out *OFFLINE* products from the result of a search. This can be done as follows currently:
```python
from eodag.plugins.crunch.filter_property import FilterProperty
dag.set_preferred_provider("peps")
prods, _ = dag.search(...)
online_prods = prods.crunch(FilterProperty({"storageStatus": "ONLINE", "operator": "eq"}))
dag.download_all(online_prods)
```
2. Add a parameter to pass as a kwarg to `EODataAccessGateway.download_all` (not sure it's relevant for `download`), e.g. `skip_offline`, that would be passed to the HTTPPlugin and, if `True`, would stop trying to download this product, which would lead to `download_all` trying to download the next product.

CGuichard:
Close #215 

Changes made to the SearchResult class (https://github.com/CS-SI/eodag/pull/359) added an easier way  to filter-out OFFLINE products.
In the original issue, that kind of code did the trick:

```python
from eodag.plugins.crunch.filter_property import FilterProperty
dag.set_preferred_provider("peps")
prods, _ = dag.search(...)
online_prods = prods.crunch(FilterProperty({"storageStatus": "ONLINE", "operator": "eq"}))
dag.download_all(online_prods)
```

Currently, with the filters attached we can do:

```python
[...]
online_prods = prods.filter_property(storageStatus="ONLINE")
[...]
```

This PR simply creates an alternative, that seems like a basic filter.

```python
[...]
online_prods = prods.filter_online()
[...]
```

Another suggestion in the issue was to add a parameter to `EODataAccessGateway.download_all`, but I think that the only goal of this method is to perform downloads. It should not modify arbitrarily the list of products passed.

### DIFF
--- a/tests/units/test_search_result.py
+++ b/tests/units/test_search_result.py
@@ -22,13 +22,34 @@ from collections import UserList
 import geojson
 from shapely.geometry.collection import GeometryCollection
 
-from tests.context import SearchResult
+from tests.context import EOProduct, SearchResult
 
 
 class TestSearchResult(unittest.TestCase):
     def setUp(self):
         super(TestSearchResult, self).setUp()
         self.search_result = SearchResult([])
+        self.search_result2 = SearchResult(
+            [
+                EOProduct(
+                    provider=None,
+                    properties={"geometry": "POINT (0 0)", "storageStatus": "ONLINE"},
+                ),
+                EOProduct(
+                    provider=None,
+                    properties={"geometry": "POINT (0 0)", "storageStatus": "OFFLINE"},
+                ),
+            ]
+        )
+
+    def test_search_result_filter_online(self):
+        """SearchResult.filter_online must only keep online results"""
+        filtered_products = self.search_result2.filter_online()
+        origin_size = len(self.search_result2)
+        filtered_size = len(filtered_products)
+        self.assertFalse(origin_size == filtered_size)
+        for product in filtered_products:
+            assert product.properties["storageStatus"] == "ONLINE"
 
     def test_search_result_geo_interface(self):
         """SearchResult must provide a FeatureCollection geo-interface"""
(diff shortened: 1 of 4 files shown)


tests/units/test_search_result.py
```diff
@@ -22,13 +22,31 @@
...
+                ),
+                EOProduct(
+                    provider=None,
+                    properties={"geometry": "POINT (0 0)", "storageStatus": "OFFLINE"},
+                ),
+            ]
+        )
+
+    def test_search_result_filter(self):
+        filtered_products = self.search_result2.filter_online()
```
sbrunato, line 46:
```suggestion
    def test_search_result_filter(self):
        """SearchResult.filter_online must only keep online results"""
        filtered_products = self.search_result2.filter_online()
```

tests/units/test_search_result.py
```diff
@@ -22,13 +22,31 @@
...
+        origin_size = len(self.search_result2)
+        filtered_size = len(filtered_products)
+        self.assertFalse(origin_size == filtered_size)
```
sbrunato, line 50:
```suggestion
        self.assertFalse(origin_size == filtered_size)
        for product in filtered_products:
            assert product.properties["storageStatus"] == "ONLINE"

```

sbrunato (changes requested):

PR #561, opened 2022-11-28
fix: handle local assets in HTTPDownload

sbrunato:
Handle local assets when some (or all) are found using `HTTPDownload` plugin


sbrunato:
> This means we can create a local STAC catalogue and download the assets ?

Yes, this will not fail. `download()` will not copy local files but return the common path of all assets. Check `test_plugins_download_http_several_local_assets` for an example

PR #681, opened 2023-03-20
feat: parse all product assets in the stac rest server json response

### LINKED ISSUE #346: Assets are not parsed to the json response in the stac rest server
When implementing a search plugin for the stac server, I return some EOProducts with an "assets" dictionary attached to them, but it is never used to format the response. Instead, just a "downloadLink" and "thumbnail", taken from stac.yml config file, are parsed. As far as I know, it would be as simple as taking this assets dictionary from "product" and copy it to the item (stac.py, __get_item_list).

Maybe I am missing something, but I have tried to configure the metadata_mapping inside the providers config file but nothing seems to work.

anesson-cs:
Fixes #346 

The complete list of stac provider product assets (not only `downloadLink` and `thumbnail`) is parsed in the stac rest server json response.

### DIFF
--- a/tests/resources/provider_responses/earth_search_search.json
+++ b/tests/resources/provider_responses/earth_search_search.json
@@ -1,0 +1,486 @@
+{
+    "type": "FeatureCollection",
+    "stac_version": "1.0.0-beta.2",
+    "stac_extensions": [],
+    "context": {
+        "page": 1,
+        "limit": 1,
+        "matched": 21830128,
+        "returned": 1
+    },
+    "numberMatched": 21830128,
+    "numberReturned": 1,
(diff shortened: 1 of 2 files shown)


tests/units/test_download_plugins.py
```diff
@@ -590,7 +590,7 @@ def run():
                 self.plugin.download(
                     self.product,
                     outputs_prefix=self.output_dir,
-                    wait=0.001 / 60,
+                    wait=0.09 / 60,
```
sbrunato:
```suggestion
                    wait=0.001 / 60,
```

sbrunato (approved):

PR #741, opened 2023-06-20
feat: server-mode parallel requests

jlahovnik:
### Description
The changes in this pull request solve the problem that only one request can be handled at the same time which could lead to long queues. Now, several different requests can be handled at the same time, e.g. a search can be done during a download.

Related to #701

### DIFF
--- a/eodag/rest/server.py
+++ b/eodag/rest/server.py
@@ -22,6 +22,7 @@ import traceback
 from contextlib import asynccontextmanager
 from distutils import dist
 from json.decoder import JSONDecodeError
+from typing import Union
 
 import pkg_resources
 from fastapi import APIRouter as FastAPIRouter
@@ -32,6 +33,7 @@ from fastapi.openapi.utils import get_openapi
 from fastapi.responses import FileResponse, ORJSONResponse
 from fastapi.types import Any, Callable, DecoratedCallable
 from starlette.exceptions import HTTPException as StarletteHTTPException
+from pydantic import BaseModel
 
 from eodag.config import load_stac_api_config
 from eodag.rest.utils import (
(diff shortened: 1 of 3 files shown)


tests/units/test_http_server.py
```diff
@@ -517,7 +517,7 @@ def test_cloud_cover_post_search(self):
             protocol="POST",
             post_data={
                 "collections": [self.tested_product_type],
-                "bbox": [0, 43, 1, 44],
+                "bbox": "0, 43, 1, 44",
```
sbrunato, line 520:
```suggestion
                "bbox": [0, 43, 1, 44],
```

eodag/rest/server.py
```diff
@@ -271,42 +273,54 @@ def stac_extension_oseo(request: Request):
     return jsonable_encoder(response)
 
 
+class SearchBody(BaseModel):
+    """
+    class which describes the body of a search request
+    """
+
+    provider: Union[str, None] = None
+    collections: List[str]
```
sbrunato, line 282:
```suggestion
    collections: Union[List[str], str]
```
Handle single collection passed in a list or as a simple str

eodag/rest/server.py
```diff
@@ -271,42 +273,54 @@ def stac_extension_oseo(request: Request):
...
+    datetime: Union[str, None] = None
+    bbox: Union[str, None] = None
```
sbrunato, line 284:
```suggestion
    bbox: Union[list, str, None] = None
```
Handle bbox passed as a list (STAC API default) or as a simple str, see https://github.com/radiantearth/stac-api-spec/tree/main/item-search#query-parameter-table

sbrunato (changes requested):

sbrunato (approved):

PR #746, opened 2023-06-23
feat: standardize output tree

### LINKED ISSUE #251: Extract SAFE products without intermediate directory
**Describe the bug**
Extraction is adding a useless subdirectory, example:
extracting my_data/S2A_MSIL1C_20180619T132231_N0206_R038_T22KGV_20180619T150922.zip leads to 
my_data/S2A_MSIL1C_20180619T132231_N0206_R038_T22KGV_20180619T150922/S2A_MSIL1C_20180619T132231_N0206_R038_T22KGV_20180619T150922.SAFE

In order to be coherent with unzipping outside eodag, e.g. `unzip S2A_MSIL1C_20180619T132231_N0206_R038_T22KGV_20180619T150922.zip`, in my opinion it would be preferable to lead to my_data/S2A_MSIL1C_20180619T132231_N0206_R038_T22KGV_20180619T150922.SAFE.

### LINKED ISSUE #266: `archive_depth` only set for PEPS and sobloo
`plugins.download.base.Download._finalize` takes care of extracting a product if required and returns the path to the product. There is a mechanism to inspect 

There is a mechanism to handle which path to return when the product is extracted:
https://github.com/CS-SI/eodag/blob/bab1b7e9c81e096d62ef6b5a2a736445c3b546d7/eodag/plugins/download/base.py#L275-L284

PEPS and sobloo have `archive_depth: 2` in their config, which means that the path returned after extraction is `outputs_prefix/product_title/subdirectory`.

I guess it needs to be set for some other providers, e.g. onda, theia, creodias (those that make use of the `HTTPDownload` plugin).

(I saw that while updating `eodag-sentinelsat`.)

anesson-cs:
fixes #251, fixes #266, and extends #251 to a general case by standardizing the output tree of all the products with the pattern `/path/to/{title}`, as voted in the pool on #446.

Here is how it works for the different kinds of products:
- zip file with useless subdirectory (e.g. `{title}.SAFE`) products: `archive_depth` parameter of affected providers is set to 2 in `providers.yml` to delete archive root and intermediate directories while extracting
- zip file with name containing unwanted extension (e.g. `.SAFE` or `.nc`) before the zip file extension (e.g. `.zip` or `.tar.gz`): we add `#remove_extension` in `title` parameter of affected providers in `providers.yml` to remove the title extension
- lone file (e.g. NetCDF or grib files) or zip file with a lone file products: a directory with the name of the product title is created to place the file in
- lone file or directory with `.zip` extension: the extension is removed and in the case of a lone file, a directory with the name of the product title is created to place the file in
- directory products downloaded with `AwsDownload` plugin: we remove the manual `{title}.SAFE` subdirectory creation when building SAFE configuration
- other kind of zip file or directory products: we keep them as they are

Notes:
- in the case of assets from `cop_ads`and `cop_cds` providers, assets are kept in their state so we may find a zipfile in a directory.


tests/test_end_to_end.py
```diff
@@ -515,7 +515,7 @@ def test_end_to_end_search_download_cop_cds(self):
         product = self.execute_search(
             *COP_CDS_SEARCH_ARGS, search_kwargs_dict=COP_CDS_SEARCH_KWARGS
         )
-        expected_filename = "{}.grib".format(product.properties["title"])
+        expected_filename = "{}".format(product.properties["title"])
```
sbrunato, line 518: test fails because file is downloaded in the following path:
```
/tmp/tmpm3fh64ma/ERA5_SL_20210101_43c1acf7bd098051d8e7bacd633482acd605f2c1.grib  
```
It should have been 
```
/tmp/tmpm3fh64ma/ERA5_SL_20210101_43c1acf7bd098051d8e7bacd633482acd605f2c1/ERA5_SL_20210101_43c1acf7bd098051d8e7bacd633482acd605f2c1.grib  
```

sbrunato (changes requested):

sbrunato (approved):

PR #750, opened 2023-06-27
feat: add provider dedl

jlahovnik:
DEDL is added as a new provider

**To be merged before June 2024**


jlahovnik:
- added authentification information to config
- urls currently set to central.hda.data, should be changed to hda.data when this url is ready

sbrunato (approved):

PR #753, opened 2023-06-30
213 fallback mechanism to search for products with other providers

jlahovnik:
Adds a fallback mechanism for the search of products. If the search request fails with the preferred providers, the provider with next highest priority will be used. The requests continue until a request is successful or until no more provider offering the desired product is found. In the latter case an error is raised.

Furthermore, the bug, that searching by ids via the /search endpoint was not working, is fixed. The parameter ids is now taken into consideration (however, only the first id is used as searching with multiple ids is currently not supported).

### DIFF
--- a/eodag/rest/utils.py
+++ b/eodag/rest/utils.py
@@ -393,6 +393,7 @@ def search_products(product_type, arguments, stac_formatted=True):
     try:
         arg_product_type = arguments.pop("product_type", None)
         unserialized = arguments.pop("unserialized", None)
+        id = arguments.pop("id", None)
 
         page, items_per_page = get_pagination_info(arguments)
         dtstart, dtend = get_datetime(arguments)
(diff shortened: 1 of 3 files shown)


sbrunato (approved):

eodag/rest/utils.py
```diff
@@ -394,6 +394,7 @@ def search_products(product_type, arguments, stac_formatted=True):
     try:
         arg_product_type = arguments.pop("product_type", None)
         unserialized = arguments.pop("unserialized", None)
+        id = arguments.pop("id", None)
```
sbrunato, line 397:
```suggestion
```

eodag/rest/utils.py
```diff
@@ -407,6 +408,7 @@ def search_products(product_type, arguments, stac_formatted=True):
             "start": dtstart,
             "end": dtend,
             "geom": geom,
+            "id": id,
```
sbrunato, line 411:
```suggestion
```

sbrunato (changes requested):

sbrunato (approved):

PR #773, opened 2023-07-26
PODAAC plugin + provider

ansotoo:
### Description
New download plugin using a session that keeps the header in redirection
New providers for PODAAC, LARC_ASDC and OB_DAAC

### Further comments
The new download plugin inherits from the HTTPDownload plugin, the only change is that it is using a SessionWithHeaderRedirection session for performing requests, instead of a simple request.
The new providers should be added to the eodag user configuration.


sbrunato:
@ansotoo the feature is not yet merged in develop because it doesn't work as is, and it needs to be integrated in eodag

AnSoTonneau:
Yes I fixed it, but I didn't commit changes yet
I will do it very soon

PR #787, opened 2023-08-09
feat: allow to download results from all pages

### LINKED ISSUE #133: handle search results pagination with a generator

anesson-cs:
Fixes #133 

After searching some products, we can now download the ones from all pages of this search by setting `exhaust` to `True` in `download_all()` method parameters.

It works by saving some elements in several objects:
- search params in `SearchResults` object to search next pages of the search
- list of crunchers in `SearchResults` object to apply them on next pages and to skip a crunch if it has already been applied on the object
- list of search params in `EODataAccessGateway` object to skip a page when it has already been requested

Then, the `search_iter_page()` is called to search products from all pages by skipping the one already found.
Finally, if there are other products, they are added to the initial `SearchResults` object.

### DIFF
--- a/eodag/api/core.py
+++ b/eodag/api/core.py
@@ -170,6 +170,9 @@ class EODataAccessGateway(object):
                     )
         self.set_locations_conf(locations_conf_path)
 
+        # record searchs done
+        self.search_kwargs_list = []
+
     def get_version(self):
         """Get eodag package version"""
         return pkg_resources.get_distribution("eodag").version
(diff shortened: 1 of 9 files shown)


sbrunato:
Following #1745, this can now be updated to use `SearchResult.next_page()`

PR #790, opened 2023-08-22
fix: search preferred provider and wekeo job error handling

### LINKED ISSUE #783: [eodag-server] setting provider parameter in a request, default this provider as preferred for future requests
**Describe the bug**

You first request a product type from a specific provider to EODAG Server.
This provider is set as the preferred one for all future requests to this EODAG Server instance.

Setting the provider parameter in requests should not impact the preferred provider defined in the configuration.

**Code To Reproduce**
We assume the default provider is not set to WEkEO but another one like PEPS

1. Request using the default provider
  ```shell
  curl http://localhost:8000/search?collections=S1_SAR_GRD
  ```

3. Request specifying a custom provider
  ```shell
  curl http://localhost:8000/search?collections=S1_SAR_GRD&provider=wekeo
  ```

4. Request without specifying a provider
  ```shell
  curl http://localhost:8000/search?collections=S1_SAR_GRD
  ```

The 3rd request use the same provider as the 2nd (a.k.a WEkEO)

**Additional context**

This behavior is caused by

```python
# https://github.com/CS-SI/eodag/blob/develop/eodag/rest/stac.py#L86-L87

        if self.provider and self.eodag_api.get_preferred_provider() != self.provider:
            self.eodag_api.set_preferred_provider(self.provider)
```

It has been introduced in commit: https://github.com/CS-SI/eodag/commit/05b24fc692cc5b55400ed8339b638b1eafe4d0e9

jlahovnik:
Contains the following changes:

- The preferred provider is not changed anymore when a search or download is executed. If a provider is given in the request, it will be used (if it offers the selected product type) but the preferred provider is not changed. If no provider is given, the configured preferred provider is used.

Fixes #783 

- an error is raised when the creation of a data request job fails for wekeo


eodag/plugins/search/data_request_search.py
```diff
@@ -113,6 +113,7 @@ def _create_data_request(self, product_type, eodag_product_type, **kwargs):
                     str(e),
                     request_job.text,
                 )
+                raise
```
sbrunato, line 116: Use `eodag.utils.exceptions.RequestError` instead of `logger.error() ; raise`

sbrunato (changes requested):
Please also add a test to check the new behavior with preferred providers

eodag/plugins/search/data_request_search.py
sbrunato, line 116: Use `eodag.utils.exceptions.RequestError` instead of `logger.error() ; raise`

sbrunato (changes requested):

eodag/plugins/search/data_request_search.py
jlahovnik: Done.

sbrunato (approved):

PR #791, opened 2023-08-23
fix: forbidden and unauthorized errors in token auth

jlahovnik:
- retries are added to the request to fetch the token in case certain errors occur (mainly intended for random unauthorized errors from wekeo)
- the token is now correctly updated in the authorization header when the previous token is expired and a new token is returned by the provider

### DIFF
--- a/eodag/plugins/authentication/token.py
+++ b/eodag/plugins/authentication/token.py
@@ -15,18 +15,27 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import logging
 
 import requests
 from requests import RequestException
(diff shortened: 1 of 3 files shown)


eodag/plugins/authentication/token.py
```diff
@@ -76,6 +90,11 @@ def authenticate(self):
             response.raise_for_status()
         except RequestException as e:
             response_text = getattr(e.response, "text", "").strip()
+            logger.error(
+                f"Could no get authentication token: {str(e)}, {response_text}"
+            )
+            if e.response and e.response.status_code == 401:
+                raise Exception  # do not forward unauthorized from provider to user
```
sbrunato, line 97:
```suggestion
```

sbrunato (changes requested):

PR #820, opened 2023-09-11
760 split atmostpheric products by variables and time

jlahovnik:
- splitting implemented for ecmwf plugin, other plugins will follow
- products can be split by year or by month (currently one month configured for ecmwf)
- one asset per selected/available parameter is created
- dummy constraint file for TIGGE_CF_SFC collection (only product type configured for ecmwf in EODAG) added until real constraints available, the dummy file covers a large number of variables from 07/2014 to 08/2023
- in ecmwf intervals are defined with dates, preparation for intervals selected with day/month/year checkboxes has also been done


jlahovnik:
merged with PR https://github.com/CS-SI/eodag/pull/826

PR #826, opened 2023-09-14
Split atmostpheric products by variables and time for cop_cds and ecmwf

jlahovnik:
- adds the "split configuration" for the cop cds and ecmwf provider
- splitting is added for the cds and ecmwf api plugin
- products can be split by year or by month (currently one month configured for both providers)
- one asset per selected/available parameter is created
- handling of constraints with either select boxes for year, month, day or date ranges available
- constraint file for ERA5_SL collection is added (only product type configured for cop_cds); currently no constraints for TIGGE_CF_SFC (only collection configured for ecmwf) available


jlahovnik:
changes in https://github.com/CS-SI/eodag/pull/889

PR #833, opened 2023-09-18
Split atmostpheric products by variables and time cop_ads

jlahovnik:
- configure splitting of products for provider cop_ads
- remove variables that are not available for ads endpoint from config


jlahovnik:
https://github.com/CS-SI/eodag/pull/889

PR #835, opened 2023-09-18
feat: return all available providers for collection

jlahovnik:
In server mode all available providers of a collection sorted by priority are no returned at the collections endpoint (GET /collections, GET /collections/id)


sbrunato (changes requested):
if provider is specified by the user, search fallback will be disabled. So in this case, we should not return the full list, but only this provider

jlahovnik:
> if provider is specified by the user, search fallback will be disabled. So in this case, we should not return the full list, but only this provider

changed

sbrunato (approved):

PR #856, opened 2023-10-03
fix: return only user requested DataRequestSearch item if id is present

alambare-csgroup:
Return the single requested item if the requested item id is found in the results.

### DIFF
--- a/eodag/plugins/search/data_request_search.py
+++ b/eodag/plugins/search/data_request_search.py
@@ -317,10 +317,8 @@ class DataRequestSearch(Search):
         """Build EOProducts from provider results"""
         results_entry = self.config.results_entry
         results = result_data[results_entry]
-        normalize_remaining_count = len(results)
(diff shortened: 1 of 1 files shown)


eodag/plugins/search/data_request_search.py
```diff
@@ -317,10 +317,15 @@ def _convert_result_data(
...
         results_entry = self.config.results_entry
         results = result_data[results_entry]
-        normalize_remaining_count = len(results)
+        # postprocess filtering needed when provider does not natively offer
+        # filtering by id
+        if "id" in kwargs:
+            results = [
+                r for r in results if r["productInfo"]["product"] == kwargs["id"]
+            ] or results
+
```
sbrunato, line 326: these test & filtering should be done after normalizing by checking `products[].properties["id"]`

sbrunato (changes requested):

PR #862, opened 2023-10-04
Split atmostpheric products by variables and time wekeo

jlahovnik:
- implement splitting of atmospheric products for data request search
- configure split parameters for ERA5 products
- request param to make splitting configurable (default False)
- implement pagination for splitted products


jlahovnik:
changes in PR https://github.com/CS-SI/eodag/pull/889

Issue #849, opened 2023-09-29
provider in item metadata


sbrunato:
Add `provider` information in server-mode response `item` metadata 
See https://github.com/radiantearth/stac-spec/blob/master/item-spec/common-metadata.md#provider-object

PR #879, opened 2023-10-12
feat: add providers metadata to items endpoint

bjgiraudon:
### Description

This PR adds STAC `providers` metadata in `/collections/{collection_id}/items` and `/collections/{collection_id}/items/{item_id}` endpoints, fixing #849 
It was previously implemented in collections, by PR #835


sbrunato (changes requested):
please move `providers` to properties and set its value as a list instead of a dict

sbrunato (approved):

PR #916, opened 2023-11-08
850 avoid search by id before every server mode download

jlahovnik:
- download info is stored in instance of EODataAccessGateway with key composed of product type, provider and id
- if download is attempted for product where id is not in download info and search by id is available for provider and product type, it will be executed; if search by id is not possible an error will be raised to do a search


jlahovnik:
updated based on input from @alambare-csgroup

eodag/api/core.py
```diff
@@ -185,6 +185,7 @@ def __init__(
                         os.path.join(self.conf_dir, "shp"),
                     )
         self.set_locations_conf(locations_conf_path)
+        self.download_info = {}
```
sbrunato, line 188: Please store full SearchResult objects in a `_search_results_cache = collections.dequeue()` having `maxlen` default to 100 (stored in a new constant) and customizable using an environment variable like `EODAG_SEARCH_RESULTS_CACHE_MAXLEN`.

Please also put `_search_results_cache` in `rest.utils` as it only concerns server-mode downloads.

To find a matching product, loop over the cache starting with the most recent result. (check at first only provider and product type of the first product of each search result)

sbrunato (changes requested):

jlahovnik:
closed because change requests were implemented on new branch -> new PR: [https://github.com/CS-SI/eodag/pull/1012](https://github.com/CS-SI/eodag/pull/1012)

PR #917, opened 2023-11-08
feat: implement new core method get_queryables

dalpasso:
Implements a new core method `def get_queryables(provider=None, product_type=None) -> List[str]`:

- If no `provider` and no `product_type` are passed, then returns default search method queryable parameters: `["productType", "start", "end", "geom", "locations", "id"]`
- If a `provider` is passed (but no `product_type`), then returns its queryable parameters from provider's config
- If a `product_type` is passed (but no `provider`), then return the intersection of the queryables of all the providers supporting it
- If both a `product_type` and a `provider` are passed, then returns the queryable parameters from provider's config and the queryable parameters for that product type

The following exceptions are raised:
- `UnsupportedProvider()` if the given `provider` is not found
- `UnsupportedProductType()` if the given `product_type` is not found

### DIFF
--- a/eodag/api/core.py
+++ b/eodag/api/core.py
@@ -2013,12 +2013,34 @@ class EODataAccessGateway:
         if provider is None and product_type is None:
             return {"productType", "start", "end", "geom", "locations", "id"}
 
+        if provider in self._plugins_manager.providers_config:
+            providers = [self._plugins_manager.providers_config[provider]]
+        else:
+            providers = self._plugins_manager.providers_config.values()
         queryable_properties: Set[str] = set()
-        for search_plugin in self._plugins_manager.get_search_plugins(
-            product_type, provider
-        ):
-            mapping = dict(search_plugin.config.metadata_mapping)
+        for provider_config in providers:
+            # if a product_type is given then search all the providers providing it,
+            # otherwise loop over all the providers
+            if (
+                product_type is not None
+                and product_type not in provider_config.products
+            ):
+                continue
+            if hasattr(provider_config, "search"):
+                plugin_config = provider_config.search
+            else:
+                plugin_config = provider_config.api
             # list of all provider-specific queryables
+            mapping = dict(plugin_config.metadata_mapping)
+            for key, value in mapping.items():
+                if isinstance(value, list) and "TimeFromAscendingNode" not in key:
+                    queryable_properties.add(key)
+            # list of all product_type-specific queryables
+            mapping = dict(
+                provider_config.products.get(product_type, {}).get(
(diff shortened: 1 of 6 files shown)


eodag/api/core.py
```diff
@@ -1970,3 +1970,51 @@ def get_cruncher(self, name, **options):
...
+        :returns: A set containing the EODAG queryable properties.
+        :rtype: set
+        """
+        if provider is None and product_type is None:
+            return {"productType", "start", "end", "geom", "locations", "id"}
+
+        if provider in self._plugins_manager.providers_config:
+            providers = [self._plugins_manager.providers_config[provider]]
+        else:
+            providers = self._plugins_manager.providers_config.values()
```
sbrunato, line 1992:
```suggestion
        if provider in self.providers_config:
            providers = [self.providers_config[provider]]
        else:
            providers = self.providers_config.values()
```

eodag/api/core.py
```diff
@@ -1970,3 +1970,51 @@ def get_cruncher(self, name, **options):
...
+            # otherwise loop over all the providers
+            if (
+                product_type is not None
+                and product_type not in provider_config.products
+            ):
+                continue
+            if hasattr(provider_config, "search"):
+                plugin_config = provider_config.search
+            else:
+                plugin_config = provider_config.api
```
sbrunato, line 2005:
```suggestion
            elif hasattr(provider_config, "api"):
                plugin_config = provider_config.api
            else:
                continue
```

sbrunato (changes requested):
- if given provider is not found, please raise `UnsupportedProvider()`
- if given product_type is not found, please raise `UnsupportedProductType()`
- if product type is given, but no provider, only return preferred provider queryables instead of the union of all possibles

dalpasso:
Changed the implementation:

- added a default list of queryables to all the providers
- when `product_type` is given, but no `provider`, then return the intersection of the queryables of all the providers supporting it

eodag/rest/utils.py
```diff
@@ -1016,8 +1015,14 @@ def rename_to_stac_standard(key: str) -> str:
     stac_config_properties = stac_config["item"]["properties"]
 
     for stac_property, value in stac_config_properties.items():
+        if isinstance(value, list):
+            value = value[0]
         if str(value).endswith(key):
             return stac_property
+
+    if key in OSEO_METADATA_MAPPING:
```
sbrunato, line 1023: use `get_criterias_from_metadata_mapping(stac_provider_metadata_mapping, arguments)` for that

sbrunato (changes requested):

PR #918, opened 2023-11-13
feat[server]: offline products order handling

### LINKED ISSUE #871: staging status in server mode download
When downloading an OFFLINE product in server mode, use min values for `wait` and `time` and return a 202 status until product is available

dalpasso:
When downloading an OFFLINE product in server mode:

- return HTTP Status 202 with new download link containing order status check information as query parameters
- don't retry to download the product again

Closes #871 

Also fixes missing `dataset` parameter after `BuildSearchResult` search-by-id in server mode

### DIFF
--- a/eodag/plugins/download/http.py
+++ b/eodag/plugins/download/http.py
@@ -582,6 +582,19 @@ class HTTPDownload(Download):
             )
         return stream_size
 
+    def _check_product_filename(self, product: EOProduct) -> str:
+        filename = None
+        asset_content_disposition = self.stream.headers.get("content-disposition", None)
+        if asset_content_disposition:
+            filename = cast(
+                Optional[str],
+                parse_header(asset_content_disposition).get_param("filename", None),
+            )
+        if not filename:
+            # default filename extracted from path
+            filename = os.path.basename(self.stream.url)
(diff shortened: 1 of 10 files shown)


eodag/rest/utils.py
```diff
@@ -686,7 +686,11 @@ def download_stac_item_by_id_stream(catalogs, item_id, provider=None):
                 "content-disposition": f"attachment; filename={os.path.basename(filepath_to_stream)}",
             },
         )
-
+    except NotAvailableError:
+        download_stream_dict = dict(
+            content=(i for i in range(0)),
+            status_code=202,
+        )
```
sbrunato, line 693:
```suggestion
    except NotAvailableError as e:
        return ORJSONResponse(
            status_code=202,
            content={"description": str(e)},
        )
```

sbrunato (changes requested):

eodag/rest/utils.py
dalpasso: Added in the commit https://github.com/CS-SI/eodag/commit/a35b21c653ceb890b2f5139a4fc71f4f46511882 with updated tests.

PR #932, opened 2023-11-21
feat: download assets subset

jlahovnik:
if a regex expression is passed in the parameter `asset` in the download only the assets matching this expression are downloaded (only working if the provider API returns individual assets and not a zip file); implemented for the `HTTPDownload` and `AwsDownload` plugins (s3rest plugin not used anymore after removal of mundi).

Client mode usage:
```py
path = dag.download(eo_product, asset=asset_key, **optional_download_kwargs)
# or
path = eo_product.assets[asset_key].download(**optional_download_kwargs)
```
Server mode usage:
http://127.0.0.1:5000/path/to/some/item/download/asset_key

### DIFF
--- a/docs/notebooks/api_user_guide/7_download.ipynb
+++ b/docs/notebooks/api_user_guide/7_download.ipynb
@@ -1112,16 +1112,242 @@
    ]
   },
   {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Download assets"
+   ]
+  },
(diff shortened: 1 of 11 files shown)


sbrunato (changes requested):
Please filter assets keys instead of title or href

jlahovnik:
> Please filter assets keys instead of title or href

done

sbrunato (approved):

PR #943, opened 2023-11-28
feat: add sorting feature for Creodias in client and server mode

### LINKED ISSUE #888: sort search results
Add a search parameter (also available in server mode) to ask products sorting if available on provider side.
Include configuration in `providers.yml`, `search.pagination`.

Usage should be:
```py
dag.search(productType="S2_MSI_L1C", sortBy=[("orbitNumber", "ASC"), ("startTimeFromAscendingNode", "DESC")])
```
Sort direction should be case insensitive, and one of (`ASCENDING`, `DESCENDING` ). We must only check the 3 first letters to also accept `ASC` or `DESC`. Ignore if no match.

Use https://github.com/stac-api-extensions/sort in server mode

anesson-cs:
fixes #888 

The set up of `sortBy` sorting search parameter has started:

- `sortBy` is implemented for the `QueryStringSearch` search plugin (and the ones using its `query()` method).

- `Creodias` is the only provider configured for this feature for now.

- `sortBy` can be used in client mode

- `sortBy` can be used in server mode in `GET` search request (following the standards described [here](https://github.com/stac-api-extensions/sort))


sbrunato (changes requested):
- [ ] please use `mypy` to fix inserted code
- [ ] if possible, add `sort_by_params` to `plugin.query_params` and `plugin .query_string` instead of storing it in `plugin.sort_by_params`
- [ ] remove unneeded url templates from provider configuration

eodag/plugins/search/base.py
```diff
@@ -157,3 +159,121 @@ def get_metadata_mapping(
...
+        if not sort_by_arg_tmp and getattr(self.config, "sort", {}).get(
+            "sort_by_default", None
+        ):
+            logger.info(
+                f"{self.provider} is configured with default sorting by '{sort_by_arg[0][0]}' "
+                f"in {'ascending' if sort_by_arg[0][1] == 'ASC' else 'descending'} order"
+            )
+        return sort_by_arg
+
+    def transform_sort_by_arg_for_search_request(
```
sbrunato, line 186: please rename to `build_sort_by` and return `(sort_by_qs: str, {}) or ("", sort_by_qp: dict)`. This will avoid `assert sort_by_provider_keyword is not None` at the end, and make it more easy to use in qssearch

sbrunato (changes requested):

sbrunato (changes requested):

sbrunato (approved):

PR #948, opened 2023-12-05
test: static type checking

sbrunato:
Adds `mypy` static type checking to tests through `tox`.

Related to #863 and #880


sbrunato:
duplicate of #1326

PR #986, opened 2024-01-12
923 new provider for creodias s3

jlahovnik:
- provider creodias s3 added
- new plugins for search (CreodiasS3Search) and download (CreodiasS3Download) created


sbrunato (changes requested):
Please add

- [x] `CreodiasS3Download` tests
- [x] documentation update for this new provider, including https://eodata-keymanager.creodias.eu/ to get credentials

sbrunato (changes requested):
- [ ] fix search result count using https://github.com/CS-SI/eodag/pull/929

PR #1000, opened 2024-01-19
fix[plugins/cds]: extract date from _dc_qs only if it exists

alambare-csgroup:
Do not try to extract date if it does not exist in query params


sbrunato:
#1000 ! 🎂🥳

sbrunato (approved):

PR #1008, opened 2024-02-01
feat: external enhanced product types metadata

dalpasso:
On server startup, load the external STAC collections linked in the `stacCollection` attribute of the product types. EODAG's properties are overridden just before returning the result to the client. The fields are updated according to the following rules:

- existing fields in EODAG's configuration are overridden;
- new fields in the external STAC collection are added to the result;
- the values of the field `"keywords"` are merged.

This feature is active only in server mode.

TODO:

- periodically reload the external STAC collections
- currently the STAC collection can be retrieved from the local file system or over HTTP (no authentication is supported)

### DIFF
--- a/eodag/api/core.py
+++ b/eodag/api/core.py
@@ -222,60 +222,6 @@ class EODataAccessGateway:
 
         self.fetch_external_product_types_metadata()
 
-    def load_external_product_type_metadata(
-        self, product_type_id: str, product_type_conf: Dict[str, Any]
-    ) -> Optional[Dict[str, Any]]:
-        """Loads external enhanced metadata for a given product type.
-
-        The existing product type configuration is merged with the one from the external file and
-        the result is returned.
-
-        Existing keys are not updated. Lists are merged also for existing keys.
-        Only the keys defined in `search.discover_product_types.generic_product_type_parsable_metadata`
-        from `eodag/resources/stac_provider.yml` are loaded into the configuration.
-
-        :param product_type_id: Product type ID to update with the external metadata.
-        :type product_type_id: str
-        :param product_type_conf: Product type configuration to update with the external
-            metadata.
-        :type product_type_conf: Dict[str, Any]
-        :returns: The enhanced product type dictionary, None if no external metadata is available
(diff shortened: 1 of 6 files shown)


eodag/api/core.py
```diff
@@ -562,6 +567,76 @@ def list_product_types(
         # Return the product_types sorted in lexicographic order of their ID
         return sorted(product_types, key=itemgetter("ID"))
 
+    def load_external_product_type_metadata(
```
sbrunato, line 570: do not add any new public core method for this feature

sbrunato (changes requested):
- enhanced collections files must be STAC formatted collection json files, not `EODAG_EXT_PRODUCT_TYPES_CFG_FILE`-like files
-

sbrunato (changes requested):
- [x] Please remove core methods, and drop client mode features.

We will only use this feature in server mode. 
In server mode,

- [x] fetch stac collections once, and update `eodag_api` product types conf, maybe using `core.update_product_types_list`

eodag/rest/stac.py
```diff
@@ -609,6 +611,25 @@ class StacCollection(StacCommon):
...
+    def fetch_external_stac_collections(cls, eodag_api: EODataAccessGateway) -> None:
+        """Load external STAC collections
+
+        :param eodag_api: EODAG python API instance
+        :type eodag_api: :class:`eodag.api.core.EODataAccessGateway`
+        """
+        list_product_types = eodag_api.list_product_types(fetch_providers=False)
+        for product_type in list_product_types:
+            ext_stac_collection_path = product_type.get("stacCollection")
+            if not ext_stac_collection_path:
```
sbrunato, line 627:
```suggestion
            if not ext_stac_collection_path := product_type.get("stacCollection"):
```

tests/units/test_stac_utils.py
sbrunato, line 1: please test `get_ext_stac_collection`

sbrunato (changes requested):

sbrunato (approved):

PR #1012, opened 2024-02-06
850 cache search results

### LINKED ISSUE #850: avoid search by id before every server mode download
Keep a cache in eodag-server of items got from /search, /items, or items/foo (limited amount) to avoid searching by id before every download

jlahovnik:
search result, provider and product type are stored in cache after search request and retrieved at download, per default data will be stored in memory but a redis cache can also be used

old caching mechanism for wekeo removed

### DIFF
--- a/eodag/plugins/search/data_request_search.py
+++ b/eodag/plugins/search/data_request_search.py
@@ -108,7 +108,7 @@ class DataRequestSearch(Search):
             self.config.pagination["next_page_url_key_path"] = string_to_jsonpath(
                 self.config.pagination.get("next_page_url_key_path", None)
             )
-        self.download_info: Dict[str, Any] = {}
+
         self.data_request_id = None
 
     def discover_product_types(self) -> Optional[Dict[str, Any]]:
@@ -386,15 +386,6 @@ class DataRequestSearch(Search):
             p.properties["orderLink"] = p.properties["orderLink"].replace(
                 "requestJobId", str(data_request_id)
             )
-            if self.config.products[product_type].get("storeDownloadUrl", False):
-                # store download information to retrieve it later in case search by id
-                # is not possible
-                self.download_info[p.properties["id"]] = {
-                    "requestJobId": data_request_id,
-                    "orderLink": p.properties["orderLink"],
-                    "downloadLink": p.properties["downloadLink"],
-                    "provider": self.provider,
-                }
         return products, total_items_nb
 
     def _check_uses_custom_filters(self, product_type: str) -> bool:
--- a/eodag/resources/providers.yml
+++ b/eodag/resources/providers.yml
@@ -1914,6 +1914,7 @@
     S2_MSI_L1C:
       productType: sentinel-s2-l1c
       metadata_mapping:
+        id: '$.properties."sentinel:product_id"'
         title: '$.properties."sentinel:product_id"'
         platformSerialIdentifier: '$.id.`split(_, 0, -1)`'
         polarizationMode: '$.id.`sub(/.{14}([A-Z]{2}).*/, \\1)`'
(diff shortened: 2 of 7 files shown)


eodag/rest/utils.py
```diff
@@ -699,6 +739,31 @@ def get_stac_item_by_id(
...
 
 
+def _retrieve_from_cache(
+    provider: Optional[str], product_type: str, item_id: str
+) -> Optional[EOProduct]:
+    if use_redis:
+        if provider:
+            cache_key = f"{product_type}_{provider}_{item_id}"
+        else:
+            cache_key = f"{product_type}_{item_id}"
```
alambare-csgroup, line 749: cf my previous comment

eodag/rest/utils.py
```diff
@@ -82,6 +84,22 @@
 
 eodag_api = eodag.EODataAccessGateway()
 
+max_len = int(os.getenv("EODAG_SEARCH_RESULTS_CACHE_MAXLEN", DEFAULT_DEQUE_MAX_LEN))
+cached_search_results: Deque = collections.deque(maxlen=max_len)
```
alambare-csgroup, line 88: Deque type should be defined. Example: Deque[Dict[str, Any]].
A dataclass would be more suited here. Something like:

@dataclass
class CacheRecord:
    """cache record"""
    provider: str
    product_type: str
    results: SearchResult
    
    with: cached_search_results: Deque[SearchRecord] = collections.deque(maxlen=max_len)

eodag/rest/utils.py
```diff
@@ -82,6 +84,22 @@
 
 eodag_api = eodag.EODataAccessGateway()
 
+max_len = int(os.getenv("EODAG_SEARCH_RESULTS_CACHE_MAXLEN", DEFAULT_DEQUE_MAX_LEN))
```
alambare-csgroup, line 87: replace: max_len = int(os.getenv("EODAG_SEARCH_RESULTS_CACHE_MAXLEN", DEFAULT_DEQUE_MAX_LEN))
by: max_len = int(os.getenv("EODAG_SEARCH_RESULTS_CACHE_MAXLEN", str(DEFAULT_DEQUE_MAX_LEN)))

cf: https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/invalid-envvar-default.html

eodag/rest/utils.py
```diff
@@ -454,6 +472,27 @@ def get_criterias_from_metadata_mapping(
     return criterias
 
 
+def _add_to_cache(product_type: str, provider: Optional[str], products: SearchResult):
```
alambare-csgroup, line 475: provider should be mandatory. We can take it from the first product of SearchResult and assume it is the same for all products of SearchResult

alambare (changes requested):

eodag/rest/utils.py
```diff
@@ -454,6 +472,27 @@ def get_criterias_from_metadata_mapping(
...
+        for product in products:
+            pickled_product = pickle.dumps(product)
+            if provider:
+                # for search done with provider
+                cache_key = f"{product_type}_{provider}_{product.properties['id']}"
+                redis_instance.set(cache_key, pickled_product)
+
+            # for search done without provider
+            cache_key = f"{product_type}_{product.properties['id']}"
+            redis_instance.set(cache_key, pickled_product)
```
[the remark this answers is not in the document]
jlahovnik: provider is now always added to cache key

eodag/rest/utils.py
jlahovnik: cache keys now always contain provider; if keys for several providers available and no provider is given, the highest priority provider is used

eodag/rest/utils.py
jlahovnik: done

eodag/rest/utils.py
jlahovnik: provider is still optional here but is added to cache key (provider from product used if not given)

PR #1023, opened 2024-02-14
feat: generate record hash from product_type + id

### LINKED ISSUE #994: Support downloaded products check from different providers.
**Is your feature request related to a problem? Please describe.**
According to this line in the code, the hash for downloaded product is calculated from the url of the product.
 https://github.com/CS-SI/eodag/blob/819247c9c57389db3b51d3d402b09edc5938ad53/eodag/plugins/download/base.py#L249C15-L249C15
 So when I download exact same product from two different providers, their hashes are different.  This feature cause the overwriting and re-downloading of exsiting downloaded product from another product. And this may be easily influenced by the url changes of provider websites.

**Describe the solution you'd like**
I think it would be better that the hash is calculated by consistent unique identifier (e.g. product.properties['id']) rather the url of the product.

dalpasso:
Fixes #994 

Avoid downloading the same product type from different providers.

The record hash is build from the concatenation of `product_type + "-" + id`

### DIFF
--- a/eodag/plugins/download/base.py
+++ b/eodag/plugins/download/base.py
@@ -86,9 +86,9 @@ class Download(PluginTopic):
       (e.g. 'file:///tmp/product_folder' on Linux or
       'file:///C:/Users/username/AppData/LOcal/Temp' on Windows)
     - save a *record* file in the directory ``outputs_prefix/.downloaded`` whose name
-      is built on the MD5 hash of the product's ``remote_location`` attribute
-      (``hashlib.md5(remote_location.encode("utf-8")).hexdigest()``) and whose content is
-      the product's ``remote_location`` attribute itself.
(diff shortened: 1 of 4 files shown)


eodag/plugins/download/base.py
```diff
@@ -278,6 +279,21 @@ def _prepare_download(
 
         return fs_path, record_filename
 
+    def generate_record_hash(self, product: EOProduct) -> str:
+        """Generate the record hash of the given product.
+
+        The MD5 hash is build from the product's ``product_type`` and ``properties['id']`` attributes
```
sbrunato, line 285:
```suggestion
        The MD5 hash is built from the product's ``product_type`` and ``properties['id']`` attributes
```

sbrunato (changes requested):

sbrunato (approved):

PR #1032, opened 2024-02-27
ssl verification changes in plugin configuration

diwakark89:
SSL Verification changes to plugins configuration and to the download and auth plugin code to pass the argument to whether allow verification while making request call or not. Also fixed the relevant unit test cases.


sbrunato (changes requested):
please only get `ssl_verify` through the configuration, not from methods arguments

PR #1046, opened 2024-03-04
fix: re-login in UsgsApi plugin on api file error

dalpasso:
If the file `~/.usgs` exists when EODAG is launched and it contains a revoked API key, then `usgs.api.login()` raises the exception `USGSError` (with message `AUTH_REVOKED: API key has been revoked, please logout and re-login.`). The file `~/.usgs` must be manually removed because calling `usgs.api.logout()` raises the same exception without removing it.

### DIFF
--- a/eodag/plugins/apis/usgs.py
+++ b/eodag/plugins/apis/usgs.py
@@ -18,6 +18,7 @@
 from __future__ import annotations
 
 import logging
+import os
 import shutil
 import tarfile
 import zipfile
@@ -101,7 +102,12 @@ class UsgsApi(Api):
             except USGSAuthExpiredError:
                 api.logout()
                 continue
-            except USGSError:
+            except USGSError as e:
+                if str(e).startswith("AUTH_REVOKED"):
+                    # If error code is "AUTH_REVOKED" then api.logout() fails to logout
+                    # Manually removing file with API key
+                    os.remove(api.TMPFILE)
+                    continue
                 raise AuthenticationError(
                     "Please check your USGS credentials."
                 ) from None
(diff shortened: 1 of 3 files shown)


eodag/plugins/apis/usgs.py
```diff
@@ -101,7 +102,12 @@ def authenticate(self) -> None:
...
-            except USGSError:
+            except USGSError as e:
+                if str(e).startswith("AUTH_REVOKED"):
+                    # If error code is "AUTH_REVOKED" then api.logout() fails to logout
+                    # Manually removing file with API key
+                    os.remove(api.TMPFILE)
+                    continue
                 raise AuthenticationError(
                     "Please check your USGS credentials."
                 ) from None
```
sbrunato, line 113:
```suggestion
                ) from e
```

eodag/plugins/apis/usgs.py
```diff
@@ -101,7 +102,12 @@ def authenticate(self) -> None:
             except USGSAuthExpiredError:
                 api.logout()
                 continue
-            except USGSError:
+            except USGSError as e:
+                if str(e).startswith("AUTH_REVOKED"):
+                    # If error code is "AUTH_REVOKED" then api.logout() fails to logout
+                    # Manually removing file with API key
```
sbrunato, line 108:
```suggestion
                if i == 0:
                    # `.usgs` API file key might be obsolete
                    # Remove it and try again
```

sbrunato (changes requested):

sbrunato (approved):

PR #1071, opened 2024-03-22
feat: provider groups

alambare-csgroup:
in EODAG library:
- add a `by_group` option on `core.available_providers`. If set, we return the unique list of groups instead of the list of provider names.
- update `list_product_types` to accept group name as provider names. Calling `list_product_types` with a group name set in the provider name will return the union of product types from all the providers of this group.
- update `get_search_plugin` to accept group names as provider names. Calling `get_search_plugin` with a group name as provider name will yield all the plugins belonging to the group.

in EODAG server:
- use group when set instead of provider name

> `group` is an optional field

Example:

I have 2 provider configuration in EODAG:

- earth_search
- earth_search_cog

I configure both with `group` **earth_search**. 

- [x] When I `list  the product types` for provider **earth_search**, EODAG return the merged list of product types from earth_search and earth_search_cog. 
- [x] In server mode, the returned product types display **earth_search** as provider name.

- [x] When I `search` for products with provider **earth_search**, EDOAG search on both provider using the existing fallback mechanism.

- [x] When I ask for `queryables `with provider **earth_search** the response is the intersect of queryables from earth_search and earth_search_cog. 
- [x] `available_providers` returns only one entry for both: **earth_search**.

### DIFF
--- a/eodag/rest/stac.py
+++ b/eodag/rest/stac.py
@@ -178,10 +178,10 @@ class StacCommon:
         ][0]
         return {
             "name": getattr(provider_config, "group", provider_config.name),
-            "description": provider_config.description,
-            "roles": provider_config.roles,
-            "url": provider_config.url,
-            "priority": provider_config.priority,
+            "description": getattr(provider_config, "description", None),
+            "roles": getattr(provider_config, "roles", None),
+            "url": getattr(provider_config, "url", None),
+            "priority": getattr(provider_config, "priority", None),
         }
 
 
--- a/eodag/rest/stac.py
+++ b/eodag/rest/stac.py
@@ -729,16 +729,13 @@ class StacCollection(StacCommon):
                     )
                 ]
             )
-            providers_models: List[Dict[str, Any]] = []
-            for provider in providers:
-                providers_models.append(self.get_provider_dict(provider["name"]))
 
             # parse jsonpath
             product_type_collection = jsonpath_parse_dict_items(
                 collection_model,
                 {
                     "product_type": product_type,
-                    "providers": providers_models,
+                    "providers": [self.get_provider_dict(p) for p in providers],
                 },
             )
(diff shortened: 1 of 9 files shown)


tests/units/test_core.py
```diff
@@ -2975,3 +2978,73 @@ def test_get_product_type_from_alias(self):
...
+        providers = cls.dag.providers_config
+
+        setattr(providers[cls.group[0]], "group", cls.group_name)
+        setattr(providers[cls.group[1]], "group", cls.group_name)
+
+    def test_available_providers_by_group(self) -> None:
+        """
+        The method available_providers returns only one entry for both grouped providers
+        """
+        # drop the provider names grouped as they are grouped under "testgroup" name
```
sbrunato, line 3000:
```suggestion
        # drop the grouped names from expected providers, as they are grouped under "testgroup" name
```

sbrunato (changes requested):

eodag/rest/stac.py
```diff
@@ -166,6 +165,21 @@ def as_dict(self) -> Dict[str, Any]:
...
+            p
+            for p in self.eodag_api.providers_config.values()
+            if provider in [p.name, getattr(p, "group", None)]
+        ][0]
+        return {
+            "name": getattr(provider_config, "group", provider_config.name),
+            "description": provider_config.description,
+            "roles": provider_config.roles,
+            "url": provider_config.url,
+            "priority": provider_config.priority,
```
sbrunato, line 180:
```suggestion
            "description": getattr(provider_config, "description", None),
            "roles": getattr(provider_config, "roles", None),
            "url": getattr(provider_config, "url", None),
            "priority": getattr(provider_config, "priority", None),
```

sbrunato (changes requested):

sbrunato (approved):

PR #1105, opened 2024-04-19
feat: allow local constraints files

anesson-cs:
It is possible to use a constraints file from a local file by setting `constraints_file_url` parameter of a search plugin of a provider configuration with an absolute path or a path relative to the root of the project (`/path/to/workspace/eodag`)


sbrunato (changes requested):
- [x] please add a test with a provider configured to use a local constraints file

eodag/plugins/search/base.py
```diff
@@ -369,17 +369,29 @@ def list_queryables(
...
-        for k, v in eoadag_queryables.items():
-            field_info = get_args(v)[1] if len(get_args(v)) > 1 else None
-            if not isinstance(field_info, FieldInfo):
+        for k, v in eodag_queryables.items():
+            eodag_queryable_field_info = (
+                get_args(v)[1] if len(get_args(v)) > 1 else None
+            )
+            if not isinstance(eodag_queryable_field_info, FieldInfo):
+                continue
+            if k in filters and k in queryables:
```
sbrunato, line 381:
```suggestion
            # keep default field info of eodag queryables 
            if k in filters and k in queryables:
```

sbrunato (changes requested):

PR #1138, opened 2024-05-15
feat: add refresh token to `OIDCAuthorizationCodeFlowAuth` plugin

dalpasso:
- Add support for the refresh token
- Use the expiration time of the access token and refresh token to determine if requesting a new token or using the already retrieved one

### DIFF
--- a/eodag/plugins/authentication/openid_connect.py
+++ b/eodag/plugins/authentication/openid_connect.py
@@ -46,7 +46,131 @@ if TYPE_CHECKING:
 logger = logging.getLogger("eodag.auth.openid_connect")
 
 
-class OIDCAuthorizationCodeFlowAuth(Authentication):
+class OIDCRefreshTokenBase(Authentication):
+    """OIDC refresh token base class, to be used through specific OIDC flows plugins.
+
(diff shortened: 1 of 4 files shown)


eodag/plugins/authentication/openid_connect.py
```diff
@@ -191,11 +257,37 @@ def authenticate(self) -> CodeAuthorizedAuth:
...
+            token_response.raise_for_status()
+        except requests.exceptions.Timeout as exc:
+            raise TimeOutError(exc, timeout=HTTP_REQ_TIMEOUT) from exc
+        except Exception as exc:
+            logger.error(
+                "Could not fetch access token with refresh token, executing new token request, error: %s",
+                getattr(exc.response, "text", ""),
+            )
+            return self.request_new_token()
+        return token_response.json()
```
sbrunato, line 290: In `get_token_with_refresh_token`, please:

- [ ] get `token_time` and `refresh_time` in this method, just before sending the request
- [ ] build `refresh_token_expiration_time` and `access_token_expiration_time` from this method to make comparisons at the beginning of authenticate easier

sbrunato (changes requested):

sbrunato (approved):

PR #1199, opened 2024-06-11
fix(server): single asset download with ignore_assets plugin setting

anesson-cs:
Check if the user set `asset` in download `kwargs` so the `download_link` is not automatically used if the provider has `ignore_assets` attribute set to True in its configuration.

### DIFF
--- a/eodag/plugins/download/http.py
+++ b/eodag/plugins/download/http.py
@@ -779,7 +779,7 @@ class HTTPDownload(Download):
         # download assets if exist instead of remote_location
         if len(product.assets) > 0 and (
             not getattr(self.config, "ignore_assets", False)
-            or kwargs.get("asset", None) is not None
+            or kwargs.get("asset") is not None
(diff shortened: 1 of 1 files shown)


eodag/plugins/download/http.py
```diff
@@ -777,7 +777,10 @@ def _stream_download_dict(
             raise MisconfiguredError(f"Incompatible auth plugin: {type(auth)}")
 
         # download assets if exist instead of remote_location
-        if len(product.assets) > 0 and not getattr(self.config, "ignore_assets", False):
+        if len(product.assets) > 0 and (
+            not getattr(self.config, "ignore_assets", False)
+            or kwargs.get("asset", None) is not None
```
sbrunato, line 782:
```suggestion
            or kwargs.get("asset") is not None
```

sbrunato (changes requested):

PR #1257, opened 2024-07-11
fix(plugins): raise error if no data available on AwsDownload

alambare:
Handle correctly situations with aws plugin when no data to download is found on s3.


dalpasso:
Added test to verify that `NoMatchingProductType` is raised.

dalpasso (approved):

sbrunato (approved):

PR #1270, opened 2024-07-24
feat: create 'downloadLink' asset

### LINKED ISSUE #1042: EOProduct asset improvements
Standardize the data exposition in EO Products:
- [ ] make sure we include the "downloadLink" as an asset in EOProduct for every plugins
- [ ] downloadLink does  not reflect clearly the content of the asset. Call the "downloadLink" asset "data" instead. The description should be adapted to the content and format of the asset

Assets returned are not always retrievable using a GET request. We should also support POST method.
- [ ] For POST requests, add a method property in the asset with value set to POST. If not set we assume the request is GET

Some asset requests require a body like the datacubes datasets for example. At the moment, the body is added to the query params of the asset URL.
- [ ] if the request requires a body, add a body property to the asset with the body value

anesson-cs:
Fixes #1042 

Create a `downloadLink` asset:
- if the `downloadLink` property is got with a json path, it is kept to use it as `downloadLink` asset `href`
- if the `downloadLink` property is given with a link, we can remove it and place this link directly in the `downloadLink` asset `href`

For now, it works in library mode for 'get' download method with `HTTPDownload` plugin with provider that can contains single assets and the `downloadLink` asset at the same time".

- [ ] Update providers conf to create assets for all
- [ ] Update docs
- [ ] Create and/or update tests

### DIFF
--- a/eodag/api/product/_product.py
+++ b/eodag/api/product/_product.py
@@ -130,7 +130,7 @@ class EOProduct:
         self.provider = provider
         self.product_type = kwargs.get("productType")
         self.location = self.remote_location = properties.get("downloadLink", "")
-        self.assets = AssetsDict(self)
+        self.assets = AssetsDict(self, properties.pop("assets"))
         self.properties = {
             key: value
             for key, value in properties.items()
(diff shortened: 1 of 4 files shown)


sbrunato:
not fan of this approach. In library mode, we'd better keep working with `remote_location`, and maybe remove `properties["downloadLink"]` during `EOProduct` construction

PR #1280, opened 2024-07-30
feat: handle integers as shapefile attributes

anesson-cs:
Locations shapefile attributes are automatically converted to strings so integers like ids can be handled in search arguments:
```
locations=dict(entity_type=1234)
```

### DIFF
--- a/docs/notebooks/api_user_guide/4_search.ipynb
+++ b/docs/notebooks/api_user_guide/4_search.ipynb
@@ -1286,7 +1286,7 @@
     ")\n",
     "```\n",
     "\n",
-    "The location query (`continent=\"Europe\"`) is passed as a dictionary to the `locations` parameter. It accepts [regular expressions](https://en.wikipedia.org/wiki/Regular_expression) which can come in handy when the query field has an underlying structure (e.g. see [this tutorial](../tutos/tuto_search_location_tile.ipynb) dedicated to search for products by tile(s)).\n",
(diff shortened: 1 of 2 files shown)


docs/notebooks/api_user_guide/4_search.ipynb
```diff
@@ -1286,7 +1286,7 @@
     ")\n",
     "```\n",
     "\n",
-    "The location query (`continent=\"Europe\"`) is passed as a dictionary to the `locations` parameter. It accepts [regular expressions](https://en.wikipedia.org/wiki/Regular_expression) which can come in handy when the query field has an underlying structure (e.g. see [this tutorial](../tutos/tuto_search_location_tile.ipynb) dedicated to search for products by tile(s)).\n",
+    "The location query (`continent=\"Europe\"`) is passed as a dictionary to the `locations` parameter. It accepts [regular expressions](https://en.wikipedia.org/wiki/Regular_expression) which can come in handy when the query field has an underlying structure (e.g. see [this tutorial](../tutos/tuto_search_location_tile.ipynb) dedicated to search for products by tile(s)). As the `locations` parameter must be a string or a regex, please consider using a string even for names which are numbers like in the following example (`location_entity=\"1234\"`).\n",
```
sbrunato, line 1289:
```suggestion
    "The location query (`continent=\"Europe\"`) is passed as a dictionary to the `locations` parameter. It accepts [regular expressions](https://en.wikipedia.org/wiki/Regular_expression) which can come in handy when the query field has an underlying structure (e.g. see [this tutorial](../tutos/tuto_search_location_tile.ipynb) dedicated to search for products by tile(s)).\n",
```

sbrunato (changes requested):

sbrunato (approved):

PR #1298, opened 2024-08-14
fix(providers): remove two_passes_id_search config param

jlahovnik:
Removes `two_passes_id_search` (old `wekeo` configuration parameter) from the `providers.yml` file because it is not used anymore

### DIFF
--- a/eodag/resources/providers.yml
+++ b/eodag/resources/providers.yml
@@ -4141,7 +4141,6 @@
     need_auth: true
     auth_error_code: 401
     results_entry: 'features'
-    two_passes_id_search: true
     pagination:
       total_items_nb_key_path: '$.properties.totalResults'
       next_page_query_obj: '{{"itemsPerPage":{items_per_page},"startIndex":{skip}}}'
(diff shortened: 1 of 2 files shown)


dalpasso (commented):
`query_params_key` is still present in `config.py`

https://github.com/CS-SI/eodag/blob/0abfd6f953f3e3054cd7cf5550660eb0809d889e/eodag/config.py#L317

Is the intention of this PR to remove it only from `providers.yml`?

jlahovnik:
> `query_params_key` is still present in `config.py`
> 
> https://github.com/CS-SI/eodag/blob/0abfd6f953f3e3054cd7cf5550660eb0809d889e/eodag/config.py#L317
> 
> Is the intention of this PR to remove it only from `providers.yml`?

Good point. I remove it from config.py too.

NOTE: failing tests are not related to changes on this bracnh but to a general problem with shapely and python 3.12

dalpasso (approved):

sbrunato (approved):

PR #1331, opened 2024-10-04
feat(providers): cop_ewds as new provider

jlahovnik:
- adds a new provider `cop_ewds` ([CEMS Early Warning Data Store](https://ewds.climate.copernicus.eu/)) with similar configuration as `cop_cds` and `cop_ads`
- configuration for all available product types added
- GLOFAS, EFAS and FIRE_HISTORICAL product types have been removed from `cop_cds` because they have been moved to EWDS
- those product types have also been removed from `wekeo` because they are currently not available there


sbrunato (dismissed):
please add end-to-end tests

sbrunato (approved):

sbrunato (approved):

PR #1336, opened 2024-10-08
fix(plugins): better error handling for cop_marine

jlahovnik:
improves the error handling at the request to fetch the product type information in `CopMarineSearch`:
- only use `UnsupportedProductType` exception in case of `NotFound` error code
- otherwise `RequestError` will be raised

### DIFF
--- a/eodag/plugins/search/cop_marine.py
+++ b/eodag/plugins/search/cop_marine.py
@@ -145,7 +145,9 @@ class CopMarineSearch(StaticStacSearch):
                 logger.error("product %s not found", product_type)
                 raise UnsupportedProductType(product_type)
             logger.error("data for product %s could not be fetched", product_type)
-            raise RequestError.from_error(exc, f"data for product {product_type} could not be fetched") from exc
+            raise RequestError.from_error(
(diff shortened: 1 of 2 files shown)


eodag/plugins/search/cop_marine.py
```diff
@@ -140,9 +140,12 @@ def _get_product_type_info(
...
         try:
             collection_data = requests.get(collection_url).json()
-        except requests.RequestException:
+        except requests.RequestException as exc:
+            if exc.errno == 404:
+                logger.error("product %s not found", product_type)
+                raise UnsupportedProductType(product_type)
             logger.error("data for product %s could not be fetched", product_type)
-            raise UnsupportedProductType(product_type)
+            raise
```
sbrunato, line 148: better use `eodag.exceptions.RequestError`

sbrunato (changes requested):

eodag/plugins/search/cop_marine.py
jlahovnik: changed

sbrunato (dismissed):

sbrunato (approved):

PR #1338, opened 2024-10-10
fix(plugins): raise error when http download order fails

### LINKED ISSUE #1337: Raise an error when a request to a backend provider fails
When a request to a provider such as cop_ads fails because of a malformed request (e.g. providing a date range with an end date before the start date), eodag logs the error, which may be unnotified by the user. Instead it should raise an error.

amarandon:
Fix #1337 

When a `HttpDownload.order_download()` fails, the error is logged but this can easily go unnoticed by the user if logs aren't enabled. The purpose of this pull request is to raise an exception instead. 

However, this exception has a `status_code` attribute set to None, which makes the rest server crash when this status code is compared with integer values:
```
  File "/home/al/venvs/dedl_lab/lib/python3.12/site-packages/starlette/responses.py", line 66, in init_headers
    and not (self.status_code < 200 or self.status_code in (204, 304))
             ^^^^^^^^^^^^^^^^^^^^^^
TypeError: '<' not supported between instances of 'NoneType' and 'int'
```

This pull request provides a fix for that by ensuring that the status code defaults to 500.

### DIFF
--- a/tests/units/test_download_plugins.py
+++ b/tests/units/test_download_plugins.py
@@ -28,12 +28,14 @@ from tempfile import NamedTemporaryFile, TemporaryDirectory, gettempdir
 from typing import Any, Callable, Dict
 from unittest import mock
 
+import pytest
 import responses
 import yaml
+from requests import RequestException
 
 from eodag.api.product.metadata_mapping import DEFAULT_METADATA_MAPPING
 from eodag.utils import ProgressCallback
-from eodag.utils.exceptions import DownloadError, NoMatchingProductType
+from eodag.utils.exceptions import DownloadError, NoMatchingProductType, RequestError
 from tests import TEST_RESOURCES_PATH
 from tests.context import (
     DEFAULT_STREAM_REQUESTS_TIMEOUT,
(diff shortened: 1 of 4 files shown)


tests/units/test_http_server.py
```diff
@@ -1150,6 +1151,34 @@ def test_download_item_from_collection_no_stream(
...
+        # attribute set to None
+        exception = RequestError()
+        exception.status_code = None
+        mock_stream_download.side_effect = exception
+
+        response = self._request_valid_raw(
+            f"collections/{self.tested_product_type}/items/foo/download?provider=peps",
+            expected_status_code=500,
+        )
+        assert "RequestError" in response.text
```
jlahovnik, line 1180: We usually use self.assertIn for tests like this.

tests/units/test_download_plugins.py
```diff
@@ -1157,6 +1159,38 @@ def test_plugins_download_http_order_get(self, mock_request):
...
+        plugin = self.get_download_plugin(self.product)
+        self.product.properties["downloadLink"] = "https://peps.cnes.fr/dummy"
+        self.product.properties["orderLink"] = "http://somewhere/order"
+
+        auth_plugin = self.get_auth_plugin(plugin, self.product)
+        auth_plugin.config.credentials = {"username": "foo", "password": "bar"}
+        auth = auth_plugin.authenticate()
+
+        # Verify that a RequestError is raised
+        with pytest.raises(RequestError):
```
amarandon, line 1182: @jlahovnik If we have a policy to use unittest rather than pytest to write tests, maybe I should have used `self.assertRaises` here?

jlahovnik: Yes, I didn't see that.

sbrunato (approved):

Issue #1339, opened 2024-10-10
Include error from backend service when searching with invalid params


amarandon:
When making an invalid search request, we get an exception that isn't as descriptive as it should. For example, consider this code:
```py
from eodag import EODataAccessGateway, setup_logging
setup_logging(3)

dag = EODataAccessGateway()

dag.set_preferred_provider("dedl")

search_criteria = {
        "productType": "CAMS_EAC4",
        "start": "2024-09-01",
        "end": "2024-08-01",  # *** BEFORE START DATE ***
        "geom": {"lonmin": 1, "latmin": 43, "lonmax": 2, "latmax": 44},
        "count": True,
        "provider": "dedl",
        "raise_errors": True,
}

dag.search(**search_criteria)
```
The exception we get is:
```
eodag.utils.exceptions.RequestError: ('Skipping error while searching for dedl StacSearch instance', 
'400 Client Error: Bad Request for url: https://hda.data.destination-earth.eu/stac/search')
```
But since in this case we have enabled verbose logging, we can also see this in the console outpout:
```
2024-10-11 10:28:34,207 eodag.search.qssearch            [DEBUG   ] b'{"description":"1 error(s). datetime: Value error, Invalid datetime range, must match format (begin_date, end_date)"}'
```
If the user hasn't enabled verbose logging, they won't see this message and even with logging enabled they might miss it in the flow of messages. It would be better to include it in the exception message. As a matter of fact, some logic is present in the code to do precisely that, but it is not working as intended because it does an `if` statement to check if a variable points to `None` or a `Response` object. But a `Response` object evaluate to `False` in boolean context if it doesn't have a status code 200, so the condition isn't satisfied when we have a `Response` with an error status code and the error message included in the response coming from the provider server is not included.

PR #1341, opened 2024-10-10
fix(utils): include response text in RequestError

amarandon:
Fixes issue with missing `Response` handling in `RequestError`.

This PR includes a test triggering the bug it's addressing. The issue was due to doing an `if` on a `Response` object to test the presence of that response object. But this response object evaluates to `False` if it doesn't have a status code 200. So instead we explicitly test that the response object is not `None`.

With this change, the exception triggered by the code example provided in #1339 should include the error message contained in the response from the provider:
```
eodag.utils.exceptions.RequestError: ('Skipping error while searching for dedl StacSearch instance', '{"description":"1 error(s). datetime: Value error, Invalid datetime range, must match format (begin_date, end_date)"}')
```

### DIFF
--- a/tests/units/test_utils_exceptions.py
+++ b/tests/units/test_utils_exceptions.py
@@ -1,0 +1,23 @@
+from requests.models import Response
+
+from eodag.utils.exceptions import RequestError
+
+
+def test_request_error_includes_response_text():
+    """
+    This test triggers a bug in RequestError due to the code doing an if on a response
+    object without realizing that a response object evaluates to False if it doesn't
+    have a status code 200. This resulted in error text from the provider not being
+    included in the error.
+    """
+    original_exception = Exception()
+    response = Response()
+    response.status_code = 400
+    response._content = b"*** my response text ***"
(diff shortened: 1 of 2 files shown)


tests/units/test_utils_exceptions.py
jlahovnik, line 1: Can you create a class and use self.assertIn and self.assertEquals as it is done for example in [test_utils.py](https://github.com/CS-SI/eodag/blob/develop/tests/units/test_utils.py)?

jlahovnik (changes requested):

tests/units/test_utils_exceptions.py
amarandon: I just pushed a test for this. But now I'm wondering if it was a good idea to use `pytest.raises` rather than `self.assertRaises` in the tests I wrote for https://github.com/CS-SI/eodag/pull/1338/files

jlahovnik (approved):

sbrunato (approved):

PR #1344, opened 2024-10-14
build: add python3.13 and drop python3.8 support

sbrunato:
- Adds `python3.13` support
- Drops `python3.8` support
- Adapts github actions to use `3.9` and `3.13` as min and max supported versions

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -17,10 +17,10 @@ jobs:
         with:
           # Get history and tags for SCM versioning to work
           fetch-depth: 0
-      - name: Set up Python 3.8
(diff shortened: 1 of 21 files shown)


sbrunato:
Build fails: `eodag[server]` needs `uvicorn[standard]` which depends on `httptools` and `uvloop` that cannot build on 3.13 for the moment.
See https://github.com/MagicStack/httptools/issues/104 and https://github.com/MagicStack/uvloop/issues/622

sbrunato:
tests are now failing because `python3.13` is not supported by `stdlib-list` (used in tests), see https://github.com/pypi/stdlib-list/issues/130

PR #1353, opened 2024-10-17
feat(server): add dedicated liveness endpoint

amarandon:
At the moment we query / to ensure that the rest api is alive. This results in a lot of requests which can't be filtered out from the logs because we want to log requests to / coming from real users.

- [x] add endpoint to rest API
- [x] Modify the EO Catalogue Helm chart to use the proper endpoint to check the liveness/readiness of the STAC API component: /_mgmt/ping

### DIFF
--- a/eodag/utils/__init__.py
+++ b/eodag/utils/__init__.py
@@ -138,6 +138,9 @@ DEFAULT_MAX_ITEMS_PER_PAGE = 50
 # default product-types start date
 DEFAULT_MISSION_START_DATE = "2015-01-01T00:00:00Z"
 
+# Path of the liveness endpoint
+LIVENESS_PROBE_PATH = "/_mgmt/ping"
+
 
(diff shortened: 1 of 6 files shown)


jlahovnik (approved):

amarandon:
@jlahovnik I added a commit which filters out the message referring to the liveness endpoint from the logs

eodag/utils/__init__.py
```diff
@@ -138,6 +138,9 @@
 # default product-types start date
 DEFAULT_MISSION_START_DATE = "2015-01-01T00:00:00Z"
 
+# Path of the liveness endpoint
+LIVENESS_PROBE_PATH = "/_mgmt/ping"
+
```
sbrunato, line 143: please move to `eodag/rest/utils/__init__.py`

sbrunato (changes requested):

sbrunato (approved):

PR #1370, opened 2024-10-25
fix(plugins): auth only when needed during http download

jlahovnik:
- So far if the download of one asset was failing, this was logged but no error was raised which hid problems during the download and could lead to incomplete or even empty results while the response was still 200. This is changed now so that the download is interrupted when a problem at the download of one asset occurs.
- After this change it became visible that for `planetary_computer` some assets where not downloaded because, for the same product, there are assets where authentication is required for the download and assets where authentication should not be added. 
&rarr; In `HTTPDownload`, the asset url is now compared to `matching_url` from the config to check if authentication should be added to the request.

### DIFF
--- a/eodag/plugins/download/http.py
+++ b/eodag/plugins/download/http.py
@@ -1081,12 +1081,12 @@ class HTTPDownload(Download):
         )
         ssl_verify = getattr(self.config, "ssl_verify", True)
         matching_url = (
-            product.downloader_auth.config.matching_url
+            getattr(product.downloader_auth.config, "matching_url", "")
             if product.downloader_auth
             else ""
         )
         matching_conf = (
-            product.downloader_auth.config.matching_conf
+            getattr(product.downloader_auth.config, "matching_conf", None)
             if product.downloader_auth
             else None
         )
--- a/eodag/plugins/download/http.py
+++ b/eodag/plugins/download/http.py
@@ -19,6 +19,7 @@ from __future__ import annotations
 
 import logging
 import os
+import re
 import shutil
 import tarfile
 import zipfile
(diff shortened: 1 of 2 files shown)


eodag/plugins/download/http.py
```diff
@@ -1087,11 +1093,14 @@ def get_chunks(stream: Response) -> Any:
                     f"Local asset detected. Download skipped for {asset['href']}"
                 )
                 continue
-
+            if matching_url and re.match(matching_url, asset["href"]):
+                auth_object = auth
+            else:
+                auth_object = None
```
sbrunato, line 1099: if `matching_conf` is used instead of `matching_url` (see [auth plugins doc](https://eodag.readthedocs.io/en/latest/plugins_reference/auth.html)), `auth_object` will be `None`.
You'd better always authenticate in this case

sbrunato (changes requested):

eodag/plugins/download/http.py
jlahovnik: changed

eodag/plugins/download/http.py
```diff
@@ -1079,6 +1080,16 @@ def get_chunks(stream: Response) -> Any:
             "flatten_top_dirs", getattr(self.config, "flatten_top_dirs", True)
         )
         ssl_verify = getattr(self.config, "ssl_verify", True)
+        matching_url = (
+            product.downloader_auth.config.matching_url
```
sbrunato, line 1084:
```suggestion
            getattr(product.downloader_auth, "config.matching_url", "")
```

eodag/plugins/download/http.py
```diff
@@ -1079,6 +1080,16 @@ def get_chunks(stream: Response) -> Any:
...
+            if product.downloader_auth
+            else ""
+        )
+        matching_conf = (
+            product.downloader_auth.config.matching_conf
```
sbrunato, line 1089:
```suggestion
            getattr(product.downloader_auth, "config.matching_conf", None)
```

sbrunato (changes requested):
this was failing if `matching_url` or `matching_conf` was not set

eodag/plugins/download/http.py
jlahovnik: fixed

eodag/plugins/download/http.py
jlahovnik: fixed

sbrunato (approved):

PR #1389, opened 2024-11-06
1388 some ecmwf products have invalid default parameters

amarandon:
Fix metadata configuration of some ECMWF products to match what is actually available on the data stores.


eodag/resources/product_types.yml
```diff
@@ -2414,7 +2414,7 @@ CAMS_GFE_GFAS:
   sensorType: ATMOSPHERIC
   license: proprietary
   title: CAMS global biomass burning emissions based on fire radiative power (GFAS)
-  missionStartDate: "2003-01-01T00:00:00Z"
+  missionStartDate: "2018-07-02T00:00:00Z"
```
jlahovnik, line 2417: How did you find this date? When I go to cop ads and use their download UI, the earliest date I can select is in 2020 
image

eodag/resources/providers.yml
```diff
@@ -2913,9 +2913,9 @@
           - '{$.completionTimeFromAscendingNode#get_hydrological_year}'
     SATELLITE_CARBON_DIOXIDE:
       dataset: satellite-carbon-dioxide
-      processing_level: level_2
+      processing_level: level_3
```
jlahovnik, line 2916: In their download UI there is no day, month, year selection anymore if level_3 is selected. Did you test if the date filter is still working (it is still in the metadata mapping).

eodag/resources/providers.yml
jlahovnik, line 1: Please transfer changes from cop_ads also to wekeo_ecmwf where the default parameters might be wrong as well.

eodag/resources/providers.yml
amarandon: I works for level 2 but for level 3 it has no effect. Is it an acceptable behavior?

eodag/resources/providers.yml
amarandon: Your comment applies only to the two products for which I made changes to providers.yml, that is SATELLITE_METHANE and SATELLITE_CARBON_DIOXIDE, correct?

They seem to be correct for wekeo_ecmwf, but different. Maybe then I should have used the same values for cop_cds?

amarandon: I tried testing download through wekeo_ecmwf but getting "UnsupportedProvider: wekeo_ecmwf is not (yet) supported for SATELLITE_CARBON_DIOXIDE"

Also when listing available providers:

```python
>>> dag.available_providers(product_type="SATELLITE_CARBON_DIOXIDE")
['cop_cds', 'dedl']
>>> dag.available_providers(product_type="SATELLITE_METHANE")
['cop_cds', 'dedl']
```

But both those products are listed under the `wekeo_ecmwf` section so I don't understand why we can't use this provider to access them.

amarandon: The above was due to me having commented out wekeo-related entries in my config file. Thanks @jlahovnik !

amarandon: @jlahovnik I just pushed a fix for the format of version numbers for wekeo_ecmwf and now download is working? Do you think it's OK now?

jlahovnik (approved):

sbrunato (approved):

PR #1400, opened 2024-11-14
fix(plugins): remove default HTTPDownload zip extension

amarandon:
- [x] Remove default `.zip` extensions used when a file to download extension is unknown.
- [x] In `HTTPDownload`, use the value obtained with  `_check_product_filename()` to set the downloaded file extension.
- [x] remove `output_extension` from `providers.yaml`

### DIFF
--- a/eodag/plugins/download/base.py
+++ b/eodag/plugins/download/base.py
@@ -202,7 +202,7 @@ class Download(PluginTopic):
             or getattr(self.config, "output_dir", tempfile.gettempdir())
             or tempfile.gettempdir()
         )
-        output_extension = kwargs.get("output_extension", None) or getattr(
+        output_extension = kwargs.get("output_extension") or getattr(
             self.config, "output_extension", ""
         )
 
--- a/eodag/plugins/download/base.py
+++ b/eodag/plugins/download/base.py
@@ -203,7 +203,7 @@ class Download(PluginTopic):
             or tempfile.gettempdir()
         )
         output_extension = kwargs.get("output_extension", None) or getattr(
-            self.config, "output_extension", ".zip"
+            self.config, "output_extension", ""
         )
 
         # Strong asumption made here: all products downloaded will be zip files
@@ -345,11 +345,14 @@ class Download(PluginTopic):
         )
         output_extension = kwargs.pop("output_extension", ".zip")
 
-        product_path = (
-            fs_path[: fs_path.index(output_extension)]
-            if output_extension in fs_path
-            else fs_path
-        )
+        if output_extension:
+            product_path = (
+                fs_path[: fs_path.index(output_extension)]
+                if output_extension in fs_path
+                else fs_path
+            )
+        else:
+            product_path, _ = os.path.splitext(fs_path)
         product_path_exists = os.path.exists(product_path)
         if product_path_exists and os.path.isfile(product_path):
             logger.info(
--- a/eodag/plugins/download/base.py
+++ b/eodag/plugins/download/base.py
@@ -343,16 +343,7 @@ class Download(PluginTopic):
             if delete_archive is not None
             else getattr(self.config, "delete_archive", True)
         )
-        output_extension = kwargs.pop("output_extension", ".zip")
-
-        if output_extension:
-            product_path = (
-                fs_path[: fs_path.index(output_extension)]
-                if output_extension in fs_path
-                else fs_path
-            )
-        else:
-            product_path, _ = os.path.splitext(fs_path)
+        product_path, _ = os.path.splitext(fs_path)
         product_path_exists = os.path.exists(product_path)
         if product_path_exists and os.path.isfile(product_path):
             logger.info(
--- a/eodag/plugins/download/http.py
+++ b/eodag/plugins/download/http.py
@@ -587,13 +587,6 @@ class HTTPDownload(Download):
             )
             progress_callback = ProgressCallback(disable=True)
 
-        output_extension = getattr(self.config, "products", {}).get(
-            product.product_type, {}
-        ).get("output_extension", None) or getattr(
-            self.config, "output_extension", None
-        )
-        kwargs["output_extension"] = kwargs.get("output_extension", output_extension)
-
         fs_path, record_filename = self._prepare_download(
             product,
             progress_callback=progress_callback,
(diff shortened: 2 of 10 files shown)


jlahovnik (approved):

eodag/resources/providers.yml
```diff
@@ -2482,48 +2460,29 @@
...
         extract: true
       SATELLITE_CARBON_DIOXIDE:
-        output_extension: .zip
         extract: true
       SATELLITE_FIRE_BURNED_AREA:
-        output_extension: .zip
         extract: true
       SATELLITE_METHANE:
-        output_extension: .zip
         extract: true
```
alambare, line 2462: can we have this automatically defined ? if product is zip then we extract ? at least for this provider ?

eodag/plugins/download/http.py
```diff
@@ -979,49 +946,50 @@ def _stream_download(
...
-                progress_callback.reset(total=stream_size)
+            progress_callback.reset(total=stream_size)
+
+            def iteration_wrapper():
                 for chunk in self.stream.iter_content(chunk_size=64 * 1024):
                     if chunk:
                         progress_callback(len(chunk))
                         yield chunk
 
+            return filename, iteration_wrapper
```
sbrunato, line 991: instead of changing method signature, inspire from what was done in `_stream_download_assets` to store filename

eodag/resources/providers.yml
sbrunato: > can we have this automatically defined ? if product is zip then we extract ? at least for this provider ?

yes, `extract: true` can be set in download plugin root settings

sbrunato (changes requested):

eodag/plugins/download/http.py
sbrunato: it might be a good idea to store information in a new `EOProduct.file_metadata` (being a `TypedDict` ?). 

Use the same thing for `Asset` and move `size`, `filename`, `rel_path` inside it

eodag/resources/providers.yml
sbrunato, line 1: please rebase

sbrunato (changes requested):

eodag/resources/providers.yml
amarandon: Does this require a change in the PR or should it be addressed in a different PR?

eodag/plugins/download/base.py
```diff
@@ -345,11 +345,14 @@ def _finalize(
...
-            else fs_path
-        )
+        if output_extension:
+            product_path = (
+                fs_path[: fs_path.index(output_extension)]
+                if output_extension in fs_path
+                else fs_path
+            )
+        else:
+            product_path, _ = os.path.splitext(fs_path)
```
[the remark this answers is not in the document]
amarandon: Changed in [360bda6](https://github.com/CS-SI/eodag/pull/1400/commits/360bda6eeee9896dc3d2ab1c5f7d51d3799ea0ad)

eodag/plugins/download/http.py
```diff
@@ -908,7 +875,7 @@ def _stream_download(
         auth: Optional[AuthBase] = None,
         progress_callback: Optional[ProgressCallback] = None,
         **kwargs: Unpack[DownloadConf],
-    ) -> Iterator[Any]:
+    ) -> tuple[str, Callable[[], Any]] | None:
```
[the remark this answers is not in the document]
amarandon: Addressed in [ac9d88b](https://github.com/CS-SI/eodag/pull/1400/commits/ac9d88bb2e081fde9350e5888e05f4beee30203c)

eodag/types/download_args.py
```diff
@@ -33,7 +33,7 @@ class DownloadConf(TypedDict, total=False):
     """
 
     output_dir: str
-    output_extension: str
+    output_extension: str | None
```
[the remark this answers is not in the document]
amarandon: Addressed in [ac9d88b](https://github.com/CS-SI/eodag/pull/1400/commits/ac9d88bb2e081fde9350e5888e05f4beee30203c)

eodag/resources/providers.yml
amarandon: Response: yes this needs to be fixed as part of this PR

amarandon: I don't know how to do this. Had a looked at `eodag/plugins/download/base.py` to no avail.

eodag/plugins/download/http.py
amarandon: From what I can see, within the context of `_stream_download` the product does not yet have any assets (i.e.  `product.assets` is an empty dictionary). So there is no properties attribute to which we can add the filename.

sbrunato: to make things more simple, we can simply store the needed file information as new `EOProduct` attributes, and delay the usage of `file:` STAC extension later

amarandon: Addressed in commit https://github.com/CS-SI/eodag/pull/1400/commits/d6eae625a6b1bfb74a33d7ea0f182a3bd58506e8

eodag/resources/providers.yml
amarandon: For this particular product and its sibling, the corresponding HTTPDownload plugin configuration has `extract: false`. My understanding is that this sets the default for all products of this provider and it is then overriden to `extract: true` by some products. So unless I'm mistaking, just removing the product-specific `extract: true` will cause the behavior to change. This is the only place in `providers.yml` where I have found the `extract` set on a per-product basis.

amarandon: In fact, having done a bit more testing, it turns out that the `extract: true` at the product has no effect. Since `extract: true` is the default at the download plugin level, I've removed it from both places.

eodag/plugins/download/http.py
```diff
@@ -908,7 +876,7 @@ def _stream_download(
         auth: Optional[AuthBase] = None,
         progress_callback: Optional[ProgressCallback] = None,
         **kwargs: Unpack[DownloadConf],
-    ) -> Iterator[Any]:
+    ) -> Union[Callable[[], Any], None]:
```
sbrunato, line 879: was the `iteration_wrapper` needed ? It changes the method output, and is no more correlated to `AwsDownload._stream_download` and https://eodag.readthedocs.io/en/latest/plugins_reference/download.html#download-methods-call-graph

sbrunato (approved):

PR #1431, opened 2024-12-06
feat(core): merge queryables by provider priority

alambare:
The queryables with lower priority should not overwrite queryables from higher priority plugins.

### DIFF
--- a/tests/units/test_core.py
+++ b/tests/units/test_core.py
@@ -1632,6 +1632,68 @@ class TestCore(TestCoreBase):
         self.assertEqual(queryables["property2"], "value_dedl2")
         self.assertEqual(queryables["property3"], "value_dedl3")
 
+    @mock.patch(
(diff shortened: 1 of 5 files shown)


sbrunato (dismissed):
please add test

tests/units/test_core.py
```diff
@@ -1510,7 +1509,9 @@ def test_list_queryables(
         "eodag.plugins.search.build_search_result.ECMWFSearch.discover_queryables",
         autospec=True,
     )
-    def test_list_queryables_with_constraints(self, mock_discover_queryables: Mock):
+    def test_list_queryables_with_constraints(
```
sbrunato, line 1512: please also check `additional_properties` and `additional_information` content

sbrunato (changes requested):

sbrunato (approved):

Issue #1450, opened 2024-12-17
Disabling default product_type not working for ERA5_PL via cop_cds


amarandon:
When making this request to eodag server:

http://127.0.0.1:5000/collections/ERA5_PL/queryables?provider=cop_cds&ecmwf:product_type

The default value is still present in the response:

```javascript
  "properties": {
    "ecmwf:product_type": {
      "default": [
        "reanalysis"          /*   <-- this should not be there   */
      ],
      "items": {
        "enum": [
          "ensemble_mean",
          "ensemble_members",
          "ensemble_spread",
          "reanalysis"
        ],
        "type": "string"
      },
      "title": "product_type",
      "type": "array"
    },
```

Note that it works as expected for other products, for example:

http://127.0.0.1:5000/collections/CAMS_EAC4/queryables?provider=cop_ads&ecmwf:variable

Gives this result:

```javascript
  "properties": {
    "ecmwf:variable": {
      /* No default value, as expected */
      "description": "Please, consult the product user guide in the documentation section for more information on these variables.",
      "items": {
        "enum": [
          "10m_u_component_of_wind",
          "10m_v_component_of_wind",
          "2m_dewpoint_temperature",

```

PR #1473, opened 2025-01-15
fix(plugins): remove default queryables form value

anesson-cs:
Fixes: #1450

When a queryable is called with an empty value (`""`, `None`, `[]`, etc.) in `dag.list_queryables()`, the returned default value of this queryable must be cleared in order to let other queryables depending on it list all their available values. Now, no default value is returned and the queryable appears as required.

Moreover, if a queryable was called in `dag.list_queryables()` while it was neither a default value of the provider configuration nor in the provider constraints file but it was in the provider queryables form, an error was raised. This is possible now.

`fetch_data()` method has been exposed instead of being initialized in `ECMWFSearch` class initialization to make the code cleaner and allow the mock of this method in tests.

### DIFF
--- a/eodag/plugins/search/build_search_result.py
+++ b/eodag/plugins/search/build_search_result.py
@@ -874,9 +874,6 @@ class ECMWFSearch(PostJsonSearch):
                 if fields and (comment := fields[0].get("comment")):
                     prop["description"] = comment
 
-                if d := details.get("default"):
-                    default = default or (d[0] if fields else d)
-
             if name == "area" and isinstance(default, dict):
                 default = list(default.values())
 
--- a/eodag/plugins/search/build_search_result.py
+++ b/eodag/plugins/search/build_search_result.py
@@ -573,13 +573,13 @@ class ECMWFSearch(PostJsonSearch):
             getattr(self.config, "discover_queryables", {}).get("constraints_url", ""),
             **kwargs,
         )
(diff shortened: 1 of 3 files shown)


sbrunato (changes requested):
please test this fix

eodag/plugins/search/build_search_result.py
```diff
@@ -960,6 +957,16 @@ def format_as_provider_keyword(
...
+    @functools.lru_cache()
+    def fetch_data(self, url: str) -> Any:
+        """
+        Fetches from a provider elements like constraints or forms by using the cache.
+
+        :param url: url from which the constraints can be fetched
+        :returns: cache containing the json file content fetched from the provider
+        """
+        return self._fetch_data(url)
+
```
alambare, line 969: I don't see the point of having both `fetch_data()` and `_fetch_data()` here. What is the point of a separate method ? why not put the decorator directly on `_fetch_data()`? 

The reason we want to avoid applying lru cache to a class instance is explained here:

- https://stackoverflow.com/questions/33672412/python-functools-lru-cache-with-instance-methods-release-object
- https://rednafi.com/python/lru_cache_on_methods/

Hence the  `self.fetch_data = functools.lru_cache()(self._fetch_data)`.
What could be done to keep the function  with `def` is to remove the fetch_data() from the class or use a static method.

sbrunato (approved):

PR #1475, opened 2025-01-15
fix(plugins): raise error on failed empty stream_zip

jlahovnik:
There were situations where in server mode the response at the download was empty but the response code was 200. 
This problem occurred when a zip was created and streamed and the download of the first file failed. In that case, the download was stopped, so the result was empty, but the `stream_zip` function was hiding the raised error from the user. This problem is now fixed, an error will be returned if the download fails.

### DIFF
--- a/eodag/plugins/download/http.py
+++ b/eodag/plugins/download/http.py
@@ -778,13 +778,17 @@ class HTTPDownload(Download):
                     )
 
                 else:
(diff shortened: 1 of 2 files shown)


jlahovnik:
> Please change the docstrings of the test.

changed

anesson-cs (approved):

PR #1502, opened 2025-01-30
fix(providers): remove unnecessary metadata mappings for eumtesat_ds

jlahovnik:
Based on feedback from Eumetsat that those parameters are static, all metadata mappings "eodag -> provider" for the `eumetsat_ds` provider except for `orbitNumber`, `startTimeFromAscendingNode` and `completionTimeFromAscendingNode` are removed.
So that they are not visible in the queryables anymore. If there is no mapping "provider -> eodag", the mapping is completely removed. 
`parentIdentifier` is kept although it is also static per product type because it is required in the request to `eumetsat_ds`.

### DIFF
--- a/eodag/resources/providers.yml
+++ b/eodag/resources/providers.yml
@@ -5313,7 +5313,9 @@
         - id
         - $.properties.identifier
       # The geographic extent of the product
-      geometry: '$.geometry'
+      geometry:
+        - 'geo={geometry#to_rounded_wkt}'
+        - '$.geometry'
       defaultGeometry: 'POLYGON((180 -90, 180 90, -180 90, -180 -90, 180 -90))'
       # The url of the quicklook
       quicklook: '$.properties.links.previews[?(@.title="Quicklook")].href'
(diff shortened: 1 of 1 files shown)


anesson-cs (approved):

anesson-cs (approved):

eodag/resources/providers.yml
```diff
@@ -5334,9 +5313,7 @@
         - id
         - $.properties.identifier
       # The geographic extent of the product
-      geometry:
-        - 'geo={geometry#to_rounded_wkt}'
-        - '$.geometry'
+      geometry: '$.geometry'
```
sbrunato, line 5316: Did eumetsat tell that geometry was not queryable ? It is available from their portal UI (you can draw an AOI) and API https://data.eumetsat.int/

sbrunato (changes requested):

eodag/resources/providers.yml
jlahovnik: They just wanted end_datetime, start_datetime and sat:absolute_orbit in the queryables but geometry is indeed also queryable. For METOP datasets it does not change much because the geometries of most of the items cover almost the whole planet but for the Sentinel-3 collections it is useful, so I re-added it.

sbrunato (approved):

PR #1509, opened 2025-02-04
fix(plugins): small bugs in ecmwf queryables

jlahovnik:
fixes bugs that were causing the queryables for product type `CAMS_SOLAR_RADIATION` to fail with both providers (`cop_ads`, `wekeo_ecmwf`)

### DIFF
--- a/eodag/plugins/search/build_search_result.py
+++ b/eodag/plugins/search/build_search_result.py
@@ -887,8 +887,7 @@ class ECMWFSearch(PostJsonSearch):
             # sometimes form returns default as array instead of string
             if (
                 default
-                and "type" in prop
-                and prop["type"] == "string"
+                and prop.get("type", "") == "string"
                 and isinstance(default, list)
             ):
                 default = ",".join(default)
--- a/eodag/plugins/search/build_search_result.py
+++ b/eodag/plugins/search/build_search_result.py
@@ -592,6 +592,7 @@ class ECMWFSearch(PostJsonSearch):
                 "startTimeFromAscendingNode",
                 "completionTimeFromAscendingNode",
                 "geom",
+                "geometry",
             ):
                 formated_kwargs[key] = kwargs[key]
             else:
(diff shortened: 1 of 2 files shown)


anesson-cs (changes requested):

eodag/plugins/search/build_search_result.py
```diff
@@ -772,7 +773,10 @@ def available_values_from_constraints(
                 # we assume that if the first value is an interval, all values are intervals
                 present_values = []
                 if keyword == "date" and "/" in entry[keyword][0]:
-                    if any(is_range_in_range(x, values[0]) for x in entry[keyword]):
+                    input_range = values
```
anesson-cs, line 776: Can you add tests with `values` as `list` and as `str` please?

anesson-cs (changes requested):

eodag/plugins/search/build_search_result.py
jlahovnik: test added

anesson-cs (approved):

eodag/plugins/search/build_search_result.py
```diff
@@ -881,7 +885,11 @@ def queryables_by_form(
                 default = strip_quotes(default)
 
             # sometimes form returns default as array instead of string
-            if default and prop["type"] == "string" and isinstance(default, list):
+            if (
+                default
+                and prop.get("type", "") == "string"
```
sbrunato, line 890:
```suggestion
                and prop.get("type") == "string"
```

tests/units/test_search_plugins.py
```diff
@@ -2343,6 +2343,33 @@ def test_plugins_search_ecmwfsearch_with_custom_producttype(self):
             except Exception:
                 assert eoproduct.properties[param] == self.custom_query_params[param]
 
+    def test_plugins_search_ecmwfsearch_get_available_values_from_contraints(self):
+        constraints = [
```
sbrunato, line 2347:
```suggestion
    def test_plugins_search_ecmwfsearch_get_available_values_from_constraints(self):
        """ECMWFSearch must return available values from constraints"""
        constraints = [
```

sbrunato (changes requested):

sbrunato (approved):

PR #1514, opened 2025-02-06
fix(providers): remove eumetsat_ds duplicate product types

jlahovnik:
removes 2 product types which exist twice with different ids and adds missing missionEndDates


anesson-cs (changes requested):
Can you use this PR to refactor description of the duplicates and other `MSG` product types which are on only one line?
It would be also good to make the configuration of all this kind of products in a row.

jlahovnik:
> Can you use this PR to refactor description of the duplicates and other `MSG` product types which are on only one line? It would be also good to make the configuration of all this kind of products in a row.

Done

anesson-cs (approved):

sbrunato (approved):

PR #1515, opened 2025-02-10
feat(providers): add product types to dedl provider

anesson-cs:
Add all missing providers on the `dedl` provider configuration from [the DEDL portfolio](https://destine-data-lake-docs.data.destination-earth.eu/en/latest/dedl-discovery-and-data-access/DestinE-Data-Portfolio/DestinE-Data-Portfolio.html).

Please note the addition of  `archive_depth` parameter in this configuration to make direct access to product folder of all `dedl` product types. Furthermore, `ISIMIP` product types have been added while `dedl` still does not store them, but it should be the case very soon.


jlahovnik (changes requested):
I think the notebook docs/notebooks/api_user_guide/2_providers_products_available.ipynb has to be updated after adding all these product types

jlahovnik (approved):

sbrunato (approved):

PR #1556, opened 2025-03-10
fix: check ecmwf product id

anesson-cs:
When a search by `id` is done with a provider having the `ECMWFSearch` search plugin, the `id` is checked: if it is one of an existing job, the job is polled by calling `_order_status()` method, otherwise, an error is raised because it means this id does not exist.

This feature will help to handle search request errors on server mode.


cauriol:
NOTE: Please note that similar changes for the build_search_result.py file exist in the other PR: https://github.com/CS-SI/eodag/pull/1572
It would be good to verify them during the merge process.

PR #1573, opened 2025-03-24
fix(core):  Ensure datetime format compliance with STAC specification

cauriol:
According to the STAC specification, a full datetime value is expected instead of just a date. This fix ensures compliance by correctly handling datetime values.

### DIFF
--- a/eodag/resources/providers.yml
+++ b/eodag/resources/providers.yml
@@ -4435,7 +4435,7 @@
     metadata_mapping:
       startTimeFromAscendingNode: '{$.startTimeFromAscendingNode#to_iso_utc_datetime}'
       completionTimeFromAscendingNode:
-        - date={startTimeFromAscendingNode#to_iso_date}/{completionTimeFromAscendingNode#to_iso_date}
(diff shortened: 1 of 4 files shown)


eodag/resources/providers.yml
```diff
@@ -4433,10 +4433,10 @@
       product_type_fetch_url: null
       constraints_url: "https://s3.central.data.destination-earth.eu/swift/v1/dedt-lumi-constraints/{dataset}.json"
     metadata_mapping:
-      startTimeFromAscendingNode: '{$.startTimeFromAscendingNode#to_iso_date}'
+      startTimeFromAscendingNode: '{$.startTimeFromAscendingNode#to_iso_utc_datetime}'
       completionTimeFromAscendingNode:
-        - '{{"date": "{startTimeFromAscendingNode#to_non_separated_date}/to/{completionTimeFromAscendingNode#to_non_separated_date}"}}'
```
jlahovnik, line 4438: Why did you change this line? This is the mapping to format the request sent to the provider and changing it breaks the download.

jlahovnik (changes requested):

eodag/resources/providers.yml
cauriol: I changed it back

jlahovnik (approved):

sbrunato (approved):

Issue #1593, opened 2025-04-04
OAuth 2.0 refresh token: pass `grant_type`, `client_id` and `client_secret`


jgaucher-cs:
Hi, I'm trying to use eodag to perform oauth2 authentication and call the refresh token mechanism. 
 
Here we see the call to retrieve an access token: https://github.com/CS-SI/eodag/blob/develop/eodag/plugins/authentication/token.py#L246
 
Above, we see that credentials are passed in the request http = grant_type, client_id, client_secret, username, password: 
```python
req_kwargs[“data”] = dict(
    getattr(self.config, “req_data”, {}), **self.config.credentials
)
``` 
cf https://datatracker.ietf.org/doc/html/rfc6749#section-4.3.2
and https://datatracker.ietf.org/doc/html/rfc6749#section-2.3.1
 
We retrieve a dict which contains an access_token and a refresh_token. A few minutes later, if the access token has expired, we call the refresh token: https://github.com/CS-SI/eodag/blob/develop/eodag/plugins/authentication/token.py#L210
```python
response = session.post(
    self.config.refresh_uri,
    data={“refresh_token”: self.refresh_token},
``` 
Eodag no longer passes the username/password, which is OK and complies with https://datatracker.ietf.org/doc/html/rfc6749#section-6. But it doesn't pass the client_id/client_secret either, even though our authorization server demands it, as per https://datatracker.ietf.org/doc/html/rfc6749#section-2.3.1. So our application receives a 401 unauthorized response.

We also need to pass `grant_type=refresh_token`, see https://datatracker.ietf.org/doc/html/rfc6749#section-6
 
Would it be possible to add the client_id/secret to this request, when they are available in the req_data or credentials ? Thank you :)

PR #1597, opened 2025-04-07
fix(plugins): send client_id/client_secret with refresh_token in TokenAuth

jgaucher-cs:
Fixes https://github.com/CS-SI/eodag/issues/1593

Thank you!


sbrunato (changes requested):

sbrunato (changes requested):
Please also update `PluginConfig` in `eodag/config.py` with:
```py
    #: :class:`~eodag.plugins.authentication.token.TokenAuth` HTTP method to use
    request_method: str
```

sbrunato (approved):

PR #1629, opened 2025-04-23
fix(providers): cop_ewds metadata mapping

jlahovnik:
adds a missing mapping for provider `cop_ewds`

### DIFF
--- a/eodag/resources/providers.yml
+++ b/eodag/resources/providers.yml
@@ -6191,3 +6191,5 @@
       release_version: 5
       metadata_mapping:
         <<: *day_month_year
+    GENERIC_PRODUCT_TYPE:
+      dataset: '{productType}'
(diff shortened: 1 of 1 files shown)


anesson-cs:
This mapping could be followed by the configuration of a generic product type like it is done for other Copernicus providers:

```
    GENERIC_PRODUCT_TYPE:
      dataset: '{productType}'
```

anesson-cs (changes requested):
This mapping could be followed by the configuration of a generic product type like it is done for other Copernicus providers:

```
    GENERIC_PRODUCT_TYPE:
      dataset: '{productType}'
```

anesson-cs (approved):

eodag (approved):

sbrunato (approved):

PR #1649, opened 2025-05-06
fix(plugins): add alias to properties in cop_marine and EcmwfSearch plugins

jlahovnik:
adds the product type alias to the product properties in the `ECMWFSearch` and `CopMarineSearch` plugin; This is required to use the alias for the download metrics of `opentelemetry-instrumentation-eodag`.

### DIFF
--- a/eodag/plugins/search/build_search_result.py
+++ b/eodag/plugins/search/build_search_result.py
@@ -1176,6 +1176,7 @@ class ECMWFSearch(PostJsonSearch):
                 + "_ORDERABLE_"
                 + query_hash
             )
+            # use product_type_config as default properties
             product_type_config = getattr(self.config, "product_type_config", {})
             properties = dict(product_type_config, **properties)
 
@@ -1441,6 +1442,7 @@ class MeteoblueSearch(ECMWFSearch):
             productType=product_type,
             properties=properties,
         )
+        # use product_type_config as default properties
         product_type_config = getattr(self.config, "product_type_config", {})
         product.properties = dict(product_type_config, **product.properties)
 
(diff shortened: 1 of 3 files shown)


cauriol (approved):

sbrunato (changes requested):
- [ ] instead of copying only `"alias"` from `product_type_config` to `product.properties`, please use the same strategy as in `QueryStringSearch` (see https://github.com/CS-SI/eodag/blob/v3.4.1.post1/eodag/plugins/search/qssearch.py#L1090-L1093)
- [ ] please also add this to `MeteoblueSearch.normalize_results`

eodag/plugins/search/build_search_result.py
```diff
@@ -1176,6 +1176,8 @@ def normalize_results(
                 + "_ORDERABLE_"
                 + query_hash
             )
+            product_type_config = getattr(self.config, "product_type_config", {})
```
sbrunato, line 1179:
```suggestion
            # use product_type_config as default properties
            product_type_config = getattr(self.config, "product_type_config", {})
```

eodag/plugins/search/build_search_result.py
```diff
@@ -1439,6 +1441,8 @@ def slugify(date_str: str) -> str:
             productType=product_type,
             properties=properties,
         )
+        product_type_config = getattr(self.config, "product_type_config", {})
```
sbrunato, line 1444:
```suggestion
        # use product_type_config as default properties
        product_type_config = getattr(self.config, "product_type_config", {})
```

eodag/plugins/search/cop_marine.py
```diff
@@ -285,6 +285,8 @@ def _create_product(
             }
         }
         product = EOProduct(self.provider, properties, productType=product_type)
+        product_type_config = getattr(self.config, "product_type_config", {})
```
sbrunato, line 288:
```suggestion
        # use product_type_config as default properties
        product_type_config = getattr(self.config, "product_type_config", {})
```

sbrunato (changes requested):

PR #1654, opened 2025-05-13
feat: add configs for user-generated collections

jlahovnik:
add configs for the user-generated collections "Urban Heat Maps" and "Hydromet Extremes"


jlahovnik:
NOTE: should be merged when those product types are added on DEDL production

PR #1665, opened 2025-05-21
feat(plugins): auth token expiration margin

### LINKED ISSUE #1598: auth token expiration margin
Following https://github.com/CS-SI/eodag/pull/1590, add a new configuration parameter `token_expiration_margin` (1 minute by default) that should be used to avoid having the token being rejected because it expired between its validity check and the request.

RitejAyed:
A new configuration parameter, `token_expiration_margin` (default: 60 seconds), has been added to introduce a safety buffer that prevents authentication tokens from being rejected due to expiration occurring between the token's validity check and the actual request. 
This parameter was added to `config.py` and integrated into the `authentication` logic in both `eodag/plugins/authentication/token.py` and `eodag/plugins/authentication/openid_connect.py`, ensuring that token validity checks now consider this margin. 
Unit tests were also added to verify the correct behavior of this new logic. 

Fixes #1598

### DIFF
--- a/eodag/config.py
+++ b/eodag/config.py
@@ -628,12 +628,12 @@ class PluginConfig(yaml.YAMLObject):
     #: :class:`~eodag.plugins.authentication.token_exchange.OIDCTokenExchangeAuth`
     #: Identifies the issuer of the `subject_token`
     subject_issuer: str
-    #: :class:`~eodag.plugins.authentication.token_exchange.OIDCTokenExchangeAuth`
-    #: Audience that the token ID is intended for. :attr:`~eodag.config.PluginConfig.client_id` of the Relying Party
-    token_expiration_margin: int
     #: :class:`~eodag.plugins.authentication.openid_connect.OIDCAuthorizationCodeFlowAuth`
     #: :class:`~eodag.plugins.authentication.base.Authentication`
     #: Safety buffer to prevent token rejection from unexpected expiry between validity check and request.
+    token_expiration_margin: int
+    #: :class:`~eodag.plugins.authentication.token_exchange.OIDCTokenExchangeAuth`
+    #: Audience that the token ID is intended for. :attr:`~eodag.config.PluginConfig.client_id` of the Relying Party
     audience: str
     #: :class:`~eodag.plugins.authentication.generic.GenericAuth`
     #: which authentication method should be used
--- a/eodag/config.py
+++ b/eodag/config.py
@@ -630,6 +630,10 @@ class PluginConfig(yaml.YAMLObject):
     subject_issuer: str
     #: :class:`~eodag.plugins.authentication.token_exchange.OIDCTokenExchangeAuth`
     #: Audience that the token ID is intended for. :attr:`~eodag.config.PluginConfig.client_id` of the Relying Party
+    token_expiration_margin: int
+    #: :class:`~eodag.plugins.authentication.openid_connect.OIDCAuthorizationCodeFlowAuth`
+    #: :class:`~eodag.plugins.authentication.base.Authentication`
+    #: Safety buffer to prevent token rejection from unexpected expiry between validity check and request.
     audience: str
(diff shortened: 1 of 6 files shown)


eodag/plugins/authentication/token.py
```diff
@@ -152,9 +152,14 @@ def authenticate(self) -> AuthBase:
 
         # Use a thread lock to avoid several threads requesting the token at the same time
         with self.auth_lock:
-
+            expiration_margin = getattr(
```
jlahovnik: If the value is taken from the config, it should also be converted to a timedelta.

jlahovnik:
Please rebase the branch!

eodag/plugins/authentication/token.py
RitejAyed: I’ve fixed the issue by updating the authenticate function to convert the config value to a timedelta before use. Let me know if anything else needs adjusting!

eodag/config.py
```diff
@@ -630,6 +630,10 @@ class MetadataPreMapping(TypedDict, total=False):
     subject_issuer: str
     #: :class:`~eodag.plugins.authentication.token_exchange.OIDCTokenExchangeAuth`
     #: Audience that the token ID is intended for. :attr:`~eodag.config.PluginConfig.client_id` of the Relying Party
+    token_expiration_margin: int
+    #: :class:`~eodag.plugins.authentication.openid_connect.OIDCAuthorizationCodeFlowAuth`
+    #: :class:`~eodag.plugins.authentication.base.Authentication`
+    #: Safety buffer to prevent token rejection from unexpected expiry between validity check and request.
```
sbrunato: please move comment above the parameter

eodag/plugins/authentication/openid_connect.py
sbrunato, line 1: please include `token_expiration_margin` description in plugin docstring

eodag/plugins/authentication/token.py
sbrunato, line 1: please include `token_expiration_margin` description in plugin docstring

sbrunato (changes requested):
please move `60` value to a new constant in `utils/__init__.py`

eodag/config.py
```diff
@@ -628,6 +628,10 @@ class MetadataPreMapping(TypedDict, total=False):
     #: :class:`~eodag.plugins.authentication.token_exchange.OIDCTokenExchangeAuth`
     #: Identifies the issuer of the `subject_token`
     subject_issuer: str
+    #: :class:`~eodag.plugins.authentication.openid_connect.OIDCAuthorizationCodeFlowAuth`
+    #: :class:`~eodag.plugins.authentication.base.Authentication`
+    #: Safety buffer to prevent token rejection from unexpected expiry between validity check and request.
+    token_expiration_margin: int
```
sbrunato:
```suggestion
    #: :class:`~eodag.plugins.authentication.token.TokenAuth`
    #: :class:`~eodag.plugins.authentication.openid_connect.OIDCRefreshTokenBase`
    #: Safety buffer to prevent token rejection from unexpected expiry between validity check and request.
    token_expiration_margin: int
```

eodag/plugins/authentication/openid_connect.py
```diff
@@ -268,6 +284,8 @@ class OIDCAuthorizationCodeFlowAuth(OIDCRefreshTokenBase):
           Refers to the name of the query param to be used in the query request
         * :attr:`~eodag.config.PluginConfig.refresh_token_key` (``str``): The key pointing to
           the refresh_token in the json response to the POST request to the token server
+        * :attr:`~eodag.config.PluginConfig.token_expiration_margin` (``int``): The margin of time (in seconds)
+           before a token is considered expired. Default: 60 seconds.
```
sbrunato:
```suggestion
          before a token is considered expired. Default: 60 seconds.
```

sbrunato (approved):

PR #1666, opened 2025-05-22
fix(core): Queryables mismatch when list of possible values contains a single value

cauriol:
This pull request fixes an issue where queryables were not matching correctly.
When the list of possible values contained only a single value, the dedl provider was returning an empty list instead of the expected single-item list.

### DIFF
--- a/eodag/plugins/search/qssearch.py
+++ b/eodag/plugins/search/qssearch.py
@@ -29,7 +29,6 @@ from typing import (
     Optional,
     Sequence,
     TypedDict,
-    Union,
     cast,
     get_args,
 )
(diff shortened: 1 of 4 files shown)


eodag/plugins/search/qssearch.py
```diff
@@ -1929,7 +1931,14 @@ def discover_queryables(
...
             queryables_dict = model_fields_to_annotated(python_queryables)
-
+            if "geom" in queryables_dict:
+                queryables_dict["geom"] = Annotated[
+                    Union[str, dict[str, float], BaseGeometry],
+                    Field(
+                        None,
+                        description="Read EODAG documentation for all supported geometry format.",
+                    ),
+                ]
```
sbrunato: Update `Queryables.geom` with this information and assign it to `queryables_dict["geom"]`

eodag/types/__init__.py
```diff
@@ -149,6 +150,32 @@ def python_type_to_json(
         return None
 
 
+def resolve_type(schema: dict):
```
sbrunato: remove this method and fix `Queryables.start` & `end` from `datetime` objects to `str`

sbrunato (approved):

PR #1671, opened 2025-05-26
feat(server): Added bbox filter support for collections search

gtesoro:
Add support for a bbox filter when listing collections via /collections endpoint.


jlahovnik (changes requested):
If you enter an invalid bbox, you get an Internal Server Error instead of a useful error message (this is only logged).

PR #1688, opened 2025-06-18
build: update usgs to 0.3.6

sbrunato:
Following https://github.com/CS-SI/eodag/pull/1552, updates `usgs` to its latest version now that it has been released


kalxas:
Thank you for the update!

PR #1704, opened 2025-06-25
feat(plugins): allow multiple download or auth plugins for a single provider

jlahovnik:
Instead of a single download or authentication plugin, now a list of plugins is defined for each provider: 

- `search_auth` and `download_auth` have been removed, both are now given in a list and the parameter `required_for` (`list[str]`) is used to define if an authentication plugin is used for search, download or both. This also replaces the attribute `need_auth` of the search plugin. The parameter `required_credentials` is used to define which credentials are used for which plugin in case of multiple plugins. If it is not given, it is assumed that all available credentials from `eodag.yml` are required.
- `matching_url` and `matching_conf` have been replaced by a dictionary `match`. The parameter `href` in the dictionary corresponds to `matching_url`, the remaining parameters to `matching_conf`.
- The configurations of the existing providers have been updated.

### DIFF
--- a/eodag/config.py
+++ b/eodag/config.py
@@ -933,14 +933,22 @@ def override_config_from_env(config: dict[str, ProviderConfig]) -> None:
             # try converting env_value type from type hints
             if "list" in str(env_type):
                 # convert str to array and then cast (only working for list[str] type)
-                env_value = env_value.split(",")
-            try:
-                env_value = cast_scalar_value(env_value, env_type)
-            except TypeError:
-                logger.warning(
-                    f"Could not convert {parts[0]} value {env_value} to {env_type}"
-                )
-            mapping[parts[0]] = env_value
+                env_value_list = env_value.split(",")
+                try:
+                    env_value_list = cast_scalar_value(env_value_list, env_type)
+                except TypeError:
+                    logger.warning(
+                        f"Could not convert {parts[0]} value {env_value} to {env_type}"
+                    )
+                mapping[parts[0]] = env_value_list
+            else:
+                try:
+                    env_value = cast_scalar_value(env_value, env_type)
+                except TypeError:
+                    logger.warning(
+                        f"Could not convert {parts[0]} value {env_value} to {env_type}"
+                    )
+                mapping[parts[0]] = env_value
         else:
             new_map = mapping.setdefault(parts[0], {})
             build_mapping_from_env("__".join(parts[1:]), env_value, new_map)
--- a/eodag/plugins/manager.py
+++ b/eodag/plugins/manager.py
@@ -278,18 +278,23 @@ class PluginManager:
         if api := getattr(plugin_conf, "api", None):
             plugin_conf.api.products = plugin_conf.products
             plugin_conf.api.priority = plugin_conf.priority
-            plugin = cast(Api, self._build_plugin(product.provider, api, Api))
+            return cast(Api, self._build_plugin(product.provider, api, Api))
         elif getattr(plugin_conf, "download", None):
             plugin = next(
                 self.get_auth_or_download_plugins(
                     "download", product.provider, matching_url=matching_url
                 )
             )
+            if not isinstance(plugin, Download):
+                # basically this should not happen
+                raise MisconfiguredError(
+                    f"Invalid download plugin for provider {plugin_conf.name}"
+                )
+            return plugin
         else:
             raise MisconfiguredError(
                 f"No download plugin configured for provider {plugin_conf.name}."
             )
-        return plugin
 
     def get_auth_plugin(
         self, associated_plugin: PluginTopic, product: Optional[EOProduct] = None
@@ -307,7 +312,7 @@ class PluginManager:
         :returns: The Authentication plugin
         """
         # matching url from product to download
-        matching_url = ""
+        matching_url = None
         if product:
             matching_url = _get_matching_url_for_product(product)
         if not matching_url:
@@ -323,6 +328,11 @@ class PluginManager:
                     matching_conf=associated_plugin.config,
                 )
             )
+            if not isinstance(auth_plugin, Authentication):
+                # basically this should not happen
+                raise MisconfiguredError(
+                    f"Invalid auth plugin found: {auth_plugin.config}"
+                )
         except StopIteration:
             auth_plugin = None
         return auth_plugin
@@ -333,7 +343,7 @@ class PluginManager:
         provider: str,
         matching_url: Optional[str] = None,
         matching_conf: Optional[PluginConfig] = None,
-    ) -> Iterator[Authentication]:
+    ) -> Iterator[Union[Authentication, Download]]:
         """Build and return the authentication plugin for the given provider and
         matching conf or url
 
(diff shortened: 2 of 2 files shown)


eodag/config.py
```diff
@@ -748,9 +763,14 @@ def credentials_in_auth(auth_conf: PluginConfig) -> bool:
     :param auth_conf: Authentication plugin configuration
     :returns: True if credentials are set, else False
     """
-    return any(
```
sbrunato, line 751: move `credentials_in_auth` to `PluginConfig.has_credentials` ?

eodag/plugins/authentication/base.py
```diff
@@ -34,16 +35,27 @@ class Authentication(PluginTopic):
...
-          configuration that needs authentication and helps identifying it
+        * :attr:`~eodag.config.PluginConfig.match` (``dict[str, str]``): dict of parameters to
+          be matched with a product or a search/download plugin to find the correct auth plugin
     """
 
     def authenticate(self) -> Union[AuthBase, S3SessionKwargs]:
         """Authenticate"""
         raise NotImplementedError
 
+    def get_required_credentials(self) -> dict[str, str]:
```
sbrunato, line 46:
```suggestion
    def get_credentials(self) -> dict[str, str]:
```

eodag/plugins/authentication/base.py
```diff
@@ -34,16 +35,27 @@ class Authentication(PluginTopic):
...
+        """checks if only a subset of the credentials is required for the plugin object
+        and returns this subset; returns all credentials if not required credentials are given
+        :returns: dict of credentials
```
sbrunato, line 49:
```suggestion
        """Get plugin configuration credentials.

        Checks if only a subset of the credentials is required for the plugin object
        and returns this subset.
        Returns all credentials if no required credentials are given.

        :returns: dict of credentials
```

eodag/plugins/manager.py
```diff
@@ -55,6 +51,56 @@
 logger = logging.getLogger("eodag.plugins.manager")
 
 
+def _is_plugin_matching(
+    plugin_conf: PluginConfig,
+    matching_url: Optional[str],
+    matching_conf: Optional[PluginConfig],
+) -> bool:
+    """
+    checks if the given plugin config is matching the matching config or url
```
sbrunato, line 60:
```suggestion
    Checks if the given plugin config is matching the matching config or url

```

eodag/plugins/manager.py
```diff
@@ -55,6 +51,56 @@
 logger = logging.getLogger("eodag.plugins.manager")
 
 
+def _is_plugin_matching(
```
sbrunato, line 54: move to `PluginConfig.is_matching` ?

eodag/plugins/manager.py
```diff
@@ -260,66 +312,47 @@ def get_auth_plugin(
...
+                # basically this should not happen
+                raise MisconfiguredError(
+                    f"Invalid auth plugin found: {auth_plugin.config}"
+                )
         except StopIteration:
             auth_plugin = None
         return auth_plugin
 
-    def get_auth_plugins(
+    def get_auth_or_download_plugins(
```
sbrunato, line 340:
```suggestion
    def get_listed_plugins(
```

eodag/plugins/manager.py
```diff
@@ -260,66 +312,47 @@ def get_auth_plugin(
...
         matching_conf: Optional[PluginConfig] = None,
-    ) -> Iterator[Authentication]:
-        """Build and return the authentication plugin for the given product_type and
-        provider
+    ) -> Iterator[Union[Authentication, Download]]:
+        """Build and return the authentication plugin for the given provider and
+        matching conf or url
 
-        :param provider: The provider for which to get the authentication plugin
+        :param plugin_type: type of plugin to be found (auth or download)
```
sbrunato, line 350:
```suggestion
        :param plugin_topic: topic of plugin to be found (auth or download)
```

eodag/plugins/manager.py
```diff
@@ -260,66 +312,47 @@ def get_auth_plugin(
...
+        plugin_type: str,
         provider: str,
         matching_url: Optional[str] = None,
         matching_conf: Optional[PluginConfig] = None,
-    ) -> Iterator[Authentication]:
-        """Build and return the authentication plugin for the given product_type and
-        provider
+    ) -> Iterator[Union[Authentication, Download]]:
+        """Build and return the authentication plugin for the given provider and
+        matching conf or url
```
sbrunato, line 348:
```suggestion
        """Build and return the listed plugins for the given provider, topic and
        matching conf or url
```

eodag/api/core.py
```diff
@@ -463,7 +463,7 @@ def add_provider(
         products: dict[str, Any] = {
             GENERIC_PRODUCT_TYPE: {"productType": "{productType}"}
         },
-        download: dict[str, Any] = {"type": "HTTPDownload", "auth_error_code": 401},
+        download: list[dict[str, Any]] = [],
```
sbrunato, line 466: please use `None` as default value instead of `[]`

Please also use `None` as search default, and apply default value in code. 
See https://docs.astral.sh/ruff/rules/mutable-argument-default/

eodag/api/core.py
```diff
@@ -463,7 +463,7 @@ def add_provider(
...
         **kwargs: dict[str, Any],
```
sbrunato, line 467: please replace `**kwargs` with `auth: Optional[list[dict[str, Any]]] = None` and  `api: Optional[dict[str, Any]] = None`

eodag/api/core.py
```diff
@@ -552,36 +550,27 @@ def _prune_providers_list(self) -> None:
...
+                    if not credentials_exist:
+                        update_needed = True
+                        logger.info(
+                            "%s: provider needing auth for search has been pruned because "
+                            "no credentials could be found",
+                            provider,
+                        )
+                        self._pruned_providers_config[
+                            provider
+                        ] = self.providers_config.pop(provider)
```
sbrunato, line 571:
```suggestion
                        self._pruned_providers_config[
                            provider
                        ] = self.providers_config.pop(provider)
                        logger.info(
                            "%s: provider needing auth for search has been pruned because "
                            "no credentials could be found",
                            provider,
                        )
```

eodag/api/core.py
```diff
@@ -891,20 +880,18 @@ def discover_product_types(
...
+                if auth := self._plugins_manager.get_auth(
+                    search_plugin.provider,
+                    "search",
+                    getattr(search_plugin.config, "api_endpoint", None),
+                    search_plugin.config,
+                ):
+                    kwargs["auth"] = auth
+                else:
+                    logger.debug(
+                        f"No authentication plugin for {provider} for product types discovery found"
```
sbrunato, line 893:
```suggestion
                        "No authentication plugin for %s for product types discovery found", provider
```

eodag/plugins/manager.py
```diff
@@ -329,49 +362,77 @@ def _is_auth_plugin_matching(
...
+                            Download,
+                            self._build_plugin(plugin_provider, plugin_conf, Download),
+                        )
                 else:
                     continue
 
     def get_auth(
         self,
         provider: str,
+        required_for: str,
```
sbrunato, line 409:
```suggestion
        required_for: Literal["search", "download"],
```

eodag/plugins/manager.py
```diff
@@ -55,6 +51,56 @@
...
+    # no match
+    return False
+
+
+def _get_matching_url_for_product(eo_product: EOProduct) -> str:
+    """
+    returns the url used to find the download and auth plugin for a product
+    (currently there is only one matching_url per product, where per asset
+    downloaders are implmented, this will change)
+    :param eo_product: product for with the matching url shall be found
```
sbrunato, line 90:
```suggestion
    returns the url used to find the download and auth plugin for a product
    (currently there is only one matching_url per product, where per asset
    downloaders are implmented, this will change)

    TODO: adapt method to asset once downloadLink is handled through an asset

    :param eo_product: product for with the matching url shall be found
```

sbrunato (changes requested):

PR #1711, opened 2025-06-30
feat(plugins): new search config entry assets mapping

jlahovnik:
adds the possibility to created assets based on a mapping added to the search plugin config with the key `assets_mapping` (see related feature)

### DIFF
--- a/eodag/plugins/search/base.py
+++ b/eodag/plugins/search/base.py
@@ -460,11 +460,15 @@ class Search(PluginTopic):
                     f"asset mapping {key} skipped because no href is available"
                 )
                 continue
-            url_match = string_to_jsonpath(asset_href).find(provider_item)
-            if len(url_match) == 1:
-                url_path = url_match[0].value
+            json_url_path = string_to_jsonpath(asset_href)
+            if isinstance(json_url_path, str):
+                url_path = json_url_path
             else:
-                url_path = NOT_AVAILABLE
+                url_match = json_url_path.find(provider_item)
+                if len(url_match) == 1:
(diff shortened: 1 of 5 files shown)


eodag/plugins/search/base.py
```diff
@@ -439,3 +441,34 @@ def queryables_from_metadata_mapping(
...
+        """
+        assets_mapping = getattr(self.config, "assets_mapping", None)
+        if not assets_mapping:
+            return {}
+        assets = {}
+        for key, values in assets_mapping.items():
+            asset_href = values.get("href")
+            if not asset_href:
+                logger.warning(
+                    f"asset mapping {key} skipped because no href is available"
```
sbrunato:
```suggestion
                    "asset mapping %s skipped because no href is available", key
```

eodag/plugins/search/base.py
```diff
@@ -439,3 +441,34 @@ def queryables_from_metadata_mapping(
             ):
                 queryables[k] = v
         return queryables
+
+    def get_assets_from_mapping(self, provider_item: dict[str, Any]) -> dict[str, Any]:
+        """
+        create assets based on the assets_mapping in the provider's config
+        and an item returned by the provider
+        :param provider_item: dict of item properties returned by the provider
+        :returns: dict containing the asset metadata
```
sbrunato:
```suggestion
        Create assets based on the assets_mapping in the provider's config
        and an item returned by the provider

        :param provider_item: dict of item properties returned by the provider
        :returns: dict containing the asset metadata
```

sbrunato (approved):

PR #1715, opened 2025-07-01
fix: pagination for Fresh Data Pool

cauriol:
This PR fixes the pagination mechanism for fetching items from the Fresh Data Pool API.
Previously, passing a page parameter had no effect, as the API uses token-based pagination instead of numeric indexing. This update now follows the next token provided in the links field of each response to reach the desired page.

Changes:
- Implemented a loop to follow next tokens returned in the response, up to the requested page.
- Updated the request logic to inject the correct token in each iteration.

### DIFF
--- a/eodag/plugins/search/qssearch.py
+++ b/eodag/plugins/search/qssearch.py
@@ -832,7 +832,7 @@ class QueryStringSearch(Search):
         **kwargs: Any,
     ) -> tuple[list[str], Optional[int]]:
         """Build paginated urls"""
-        page = prep.page
+        token = prep.token
         items_per_page = prep.items_per_page
         count = prep.count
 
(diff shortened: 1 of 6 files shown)


eodag/plugins/search/qssearch.py
```diff
@@ -1722,6 +1722,19 @@ def _request(
...
+                data = response.json()
+                next_link = next(
+                    (link for link in data.get("links", []) if link["rel"] == "next"),
+                    None,
+                )
+                if next_link is None:
+                    break
+                token = next_link["body"]["token"]
+                prep.query_params["token"] = token
+                response = requests.post(
```
jlahovnik: This request should also use ssl_verify and timeout.

PR #1759, opened 2025-07-30
fix: remove instrument property if its value is None

timothey-cs:
Remove the instrument property when it's None during the result normalizing


eodag/plugins/search/qssearch.py
sbrunato, line 1: the malformed property is already present in raw provider response. Do not modify `qssearch`, this fix has to be applied on the server-mode as `dedl` uses it

sbrunato (changes requested):

PR #1770, opened 2025-08-13
refactor(plugins): dedt_lumi not-available data error message

gtesoro:
dedt_lumi respond with 400 when downloading an order that returns no data.
This PR improves the reported message to the user in such instance.

### DIFF
--- a/eodag/plugins/download/http.py
+++ b/eodag/plugins/download/http.py
@@ -334,18 +334,6 @@ class HTTPDownload(Download):
                     f"Order download status request responded with {response.status_code}"
                 )
 
-                # [CFDEDL-6623]: dedt_lumi currently responds with a 400 when no data is available.
-                # This often happens when Region of Interest polygon is not available.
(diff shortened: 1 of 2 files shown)


eodag/plugins/download/http.py
```diff
@@ -332,6 +333,19 @@ def _request(
...
                 )
+
+                # [CFDEDL-6623]: dedt_lumi currently responds with a 400 when no data is available.
+                # This often happens when Region of Interest polygon is not available.
+                # This is ad-hoc error handling for that case.
+                if (
+                    product.provider == "dedt_lumi"
+                    and "ecmwf:feature" in product.properties
+                    and response.status_code == 400
+                ):
```
sbrunato: please handle it better. We should not use such tests in the code

sbrunato (changes requested):
I could not reproduce the issue. Can you add more information on it?

sbrunato (approved):

Issue #942, opened 2023-11-23
query parameters validation


sbrunato:
In order to validate type and formatting of query parameters, please add a new `pydantic` class in `metadata_mapping.py`, describing all the possible queryables, from OSEO and also [ECMWF/MARS](https://confluence.ecmwf.int/display/UDOC/Keywords+in+MARS+and+Dissemination+requests).

Validation could be performed in `core._prepare_search`.

PR #1771, opened 2025-08-14
feat: search validation

dalpasso:
This pull request:

* adds validation of search requests in the EODAG API. The validation is enabled by default.
* the new parameter `validate` is used in `EODataAccessGateway`.
* removes the default values of the parameters from all ECMWF-like providers.

The validation of the search request is implemented in the API by adding a new optional argument `validate` to `EODataAccessGateway#search`.

Fixes https://github.com/CS-SI/eodag/issues/942

### DIFF
--- a/docs/notebooks/api_user_guide/3_search.ipynb
+++ b/docs/notebooks/api_user_guide/3_search.ipynb
@@ -5443,102 +5443,24 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "### Validate search request\n",
+    "### Validation\n",
     "\n",
-    "The parameter `validate` can be used to automatically validate a search request before sending it to the provider. The validation is enabled by default."
+    "The parameter `validate` can be used to automatically validate a search request before sending it to the provider. The validation is **enabled by default**."
    ]
   },
   {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "The following example illustrate a search request with validation disabled and the following download of the first result using invalid parameters. With ECMWF-like providers as `cop_cds`, the search is going to return a product, but the following download is going to fail."
-   ]
-  },
-  {
-   "cell_type": "code",
-   "execution_count": 11,
-   "metadata": {},
-   "outputs": [
-    {
-     "name": "stderr",
-     "output_type": "stream",
-     "text": [
-      "2025-10-06 18:48:36,135 eodag.core                       [INFO    ] Searching on provider cop_cds\n"
-     ]
-    },
-    {
-     "name": "stdout",
-     "output_type": "stream",
-     "text": [
-      "Invalid search criteria without validation: got a hand on 1 products.\n"
-     ]
-    },
-    {
-     "name": "stderr",
-     "output_type": "stream",
-     "text": [
-      "2025-10-06 18:48:36,176 eodag.download.base              [INFO    ] Download url: Not Available\n",
-      "2025-10-06 18:48:36,589 eodag.download.base              [INFO    ] Download link is not available\n"
-     ]
-    },
-    {
-     "data": {
-      "text/html": [
-       "[Retry #2] Waiting 10.588138s until next order status check (retry every 0.2' for 10')"
-      ],
-      "text/plain": [
-       "<IPython.core.display.HTML object>"
-      ]
-     },
-     "metadata": {},
-     "output_type": "display_data"
-    },
-    {
-     "name": "stderr",
-     "output_type": "stream",
-     "text": [
-      "2025-10-06 18:48:49,752 eodag.download.base              [INFO    ] Download link is not available\n"
-     ]
-    },
-    {
-     "ename": "DownloadError",
-     "evalue": "Provider cop_cds returned: None",
-     "output_type": "error",
-     "traceback": [
-      "\u001b[31m---------------------------------------------------------------------------\u001b[39m",
-      "\u001b[31mDownloadError\u001b[39m                             Traceback (most recent call last)",
-      "\u001b[36mCell\u001b[39m\u001b[36m \u001b[39m\u001b[32mIn[11]\u001b[39m\u001b[32m, line 10\u001b[39m\n\u001b[32m      8\u001b[39m products = dag.search(validate=\u001b[38;5;28;01mFalse\u001b[39;00m, **invalid_search_criteria)\n\u001b[32m      9\u001b[39m \u001b[38;5;28mprint\u001b[39m(\u001b[33mf\u001b[39m\u001b[33m\"\u001b[39m\u001b[33mInvalid search criteria without validation: got a hand on \u001b[39m\u001b[38;5;132;01m{\u001b[39;00m\u001b[38;5;28mlen\u001b[39m(products)\u001b[38;5;132;01m}\u001b[39;00m\u001b[33m products.\u001b[39m\u001b[33m\"\u001b[39m)\n\u001b[32m---> \u001b[39m\u001b[32m10\u001b[39m \u001b[43mdag\u001b[49m\u001b[43m.\u001b[49m\u001b[43mdownload\u001b[49m\u001b[43m(\u001b[49m\u001b[43mproducts\u001b[49m\u001b[43m[\u001b[49m\u001b[32;43m0\u001b[39;49m\u001b[43m]\u001b[49m\u001b[43m)\u001b[49m\n",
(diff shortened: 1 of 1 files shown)


eodag/api/core.py
```diff
@@ -25,11 +25,14 @@
...
 from importlib.metadata import version
 from importlib.resources import files as res_files
+from json import JSONDecodeError
 from operator import itemgetter
 from typing import TYPE_CHECKING, Any, Iterator, Optional, Union
+from urllib.parse import parse_qs, urlparse
 
 import geojson
 import yaml.parser
+from pydantic import ValidationError as pydanticValidationError
```
anesson-cs: We can start using PascalCase for Pydantic validation error alias: `pydanticValidationError` -> `PydanticValidationError`

eodag/plugins/search/build_search_result.py
```diff
@@ -305,11 +305,12 @@ def append_time(input_date: date, time: Optional[str]) -> datetime:
     """
     Parses a time string in format HHMM and appends it to a date.
 
-    if the time string is in format HH:MM we convert it to HHMM
+    if the time string is in format HH:MM or HH_MM we convert it to HHMM
     """
     if not time:
         time = "0000"
     time = time.replace(":", "")
+    time = time.replace("_", "")
```
anesson-cs: You can merge these two lines -> `time.replace(":", "").replace("_", "")` or `re.sub(":|_", "", time)`

eodag/api/core.py
```diff
@@ -55,10 +58,12 @@
     provider_config_init,
     share_credentials,
 )
+from eodag.plugins.download.base import Download
 from eodag.plugins.manager import PluginManager
 from eodag.plugins.search import PreparedSearch
 from eodag.plugins.search.build_search_result import MeteoblueSearch
 from eodag.plugins.search.qssearch import PostJsonSearch
+from eodag.rest.utils import format_pydantic_error
```
anesson-cs: Please duplicate this method in `eodag.utils` since the rest folder will be removed in a future eodag version.

eodag/resources/providers.yml
```diff
@@ -2272,20 +2111,8 @@
...
-      queries: [{'domain':'NEMSAUTO','gapFillDomain':null,'timeResolution':'daily','codes':[{'code':71,'level':'sfc','aggregation':'mean'}]}]
-      format: netCDF
-      units: {'temperature':'C','velocity':'km/h','length':'metric','energy':'watts'}
-      timeIntervalsAlignment:
-      geometry: {"type": "Polygon", "coordinates": [[[180, -90], [180, 90], [-180, 90], [-180, -90], [180, -90]]]}
+    NEMSGLOBAL_TCDC: {}
+    NEMSAUTO_TCDC: {}
     GENERIC_PRODUCT_TYPE:
       productType: '{productType}'
       geometry: {"type": "Polygon", "coordinates": [[[180, -90], [180, 90], [-180, 90], [-180, -90], [180, -90]]]}
```
anesson-cs, line 2254: Why did you remove `productType`, `format` and `geometry` parameters of `NEMSGLOBAL_TCDC` and `NEMSAUTO_TCDC`? It seems they are necessary to search products and to give their output format (you left `data_format` for other providers).

tests/units/test_core.py
```diff
@@ -1940,6 +1934,47 @@ def test_available_sortables(self, mock_auth_session_request):
...
+            "productType": "S1_SAR_GRD",
+            "lorem": "ipsum",
+        }
+        self.dag.search(validate_request=True, **filter)
+        mock_validate_search_request.assert_called_once()
+        args, kwargs = mock_validate_search_request.call_args
+        self.assertEqual(args[1], "peps")
+        # Some other default keyword may be added to the kwargs (e.g. geometry)
+        self.assertEqual("S1_SAR_GRD", args[2].get("productType"))
+        self.assertEqual("ipsum", args[2].get("lorem"))
```
anesson-cs: Please add a test where the validation fails.

anesson-cs (commented):
Please update documentation on the fact that there is no default values anymore for all ECMWF-like providers.

eodag/api/core.py
```diff
@@ -2491,3 +2507,51 @@ def import_stac_items(self, items_urls: list[str]) -> SearchResult:
                 results.extend(search_result)
 
         return results
+
+    def validate_search_request(self, provider: str, filter: dict[str, Any]):
```
anesson-cs: Please turn `def validate_search_request(self, provider: str, filter: dict[str, Any]):` to `def validate_search_request(self, provider: str, filter: dict[str, Any]) -> None:`

eodag/api/core.py
```diff
@@ -2491,3 +2507,51 @@ def import_stac_items(self, items_urls: list[str]) -> SearchResult:
...
+        logger.debug("Validate request")
+        try:
+            self.list_queryables(
+                provider=provider,
+                **filter,
+            ).get_model().model_validate(filter)
+        except pydanticValidationError as e:
+            raise ValidationError(format_pydantic_error(e)) from e
+
+    def validate_order_request(self, product: EOProduct):
```
anesson-cs: Please turn `def validate_order_request(self, product: EOProduct):` to `def validate_order_request(self, product: EOProduct) -> None:`

sbrunato: move method to server

eodag/api/core.py
```diff
@@ -1202,6 +1208,7 @@ def search(
         locations: Optional[dict[str, str]] = None,
         provider: Optional[str] = None,
         count: bool = False,
+        validate_request: Optional[bool] = False,
```
sbrunato:
```suggestion
        validate: Optional[bool] = False,
```
this is shorter and clear enough

eodag/api/core.py
sbrunato: move to search base plugin and rename to `validate`

sbrunato (changes requested):

eodag/api/core.py
dalpasso: I have added the field `validate` also to `SearchPostRequest`:

https://github.com/CS-SI/eodag/blob/26da0073dbfa00e4453e5297bf95d216871bddbd/eodag/rest/types/stac_search.py#L120

This shadows an attribute in parent `BaseModel`. Should I revert the name of the field to `validate_request`? Do you propose a different solution?

dalpasso: Fixed: use `validate_request` as model field and `validate` as parameter for the user in REST API and `EODataAccessGateway#search`

eodag/resources/providers.yml
dalpasso: Default values restored for the provider `meteoblue` since no constraints file is provided.

`data_format` and `download_format` are instead removed from all the ECMWF-like providers. The user must provide their values as for all other parameters for this kind of provider. If missing, the user receives a validation error before the search is sent to the remote provider.

docs/notebooks/api_user_guide/3_search.ipynb
```diff
@@ -5423,6 +5430,15 @@
     "print(f\"number_matched: {results_with_count.number_matched}\")"
    ]
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Validate search request\n",
+    "\n",
+    "The parameter `validate` can be used to automatically validate a search request before sending it to the provider. The validation is enabled by default and it's implemented by calling the pydantic method `model_validate` on the model obtained from the [queryables](5_queryables.ipynb)."
```
sbrunato:
```suggestion
    "The parameter `validate` can be used to automatically validate a search request before sending it to the provider. The validation is enabled by default."
```

docs/notebooks/api_user_guide/3_search.ipynb
sbrunato, line 1: please show examples to illustrate how validation works

eodag/api/core.py
```diff
@@ -1200,10 +1203,12 @@ def search(
             **kwargs,
         )
         if search_kwargs.get("id"):
+            # Don't validate requests by ID. "id" is not queryable.
```
sbrunato, line 1203: instead of disabling validation for id, add it as queryable when it is missing.
In the example of the doc, it is listed as queryable for `planetary_computer` but not `cop_cds`
See https://eodag--1771.org.readthedocs.build/en/1771/notebooks/api_user_guide/5_queryables.html

eodag/api/core.py
```diff
@@ -1836,6 +1843,8 @@ def _do_search(
         :param raise_errors: (optional) When an error occurs when searching, if this is set to
                              True, the error is raised
         :param kwargs: Some other criteria that will be used to do the search
+        :param validate: (optional) Set to True to validate the request query before sending it
+                         to the provider
```
sbrunato:
```suggestion
        :param validate: (optional) Set to True to validate search parameters
                         before sending the query to the provider
```

eodag/plugins/search/base.py
```diff
@@ -398,6 +400,31 @@ def list_queryables(
                 **all_queryables,
             )
 
+    def validate(
+        self, filter: dict[str, Any], auth: Optional[Union[AuthBase, S3SessionKwargs]]
```
sbrunato:
```suggestion
        self, search_params: dict[str, Any], auth: Optional[Union[AuthBase, S3SessionKwargs]]
```

eodag/plugins/search/base.py
```diff
@@ -398,6 +400,31 @@ def list_queryables(
...
+        :param filter: Arguments of the search request
+        :param auth: Authentication object
+        :raises: :class:`~eodag.utils.exceptions.ValidationError`
+        """
+        logger.debug("Validate request")
+        # attach authentication if required
+        if getattr(self.config, "need_auth", False) and auth:
+            self.auth = auth
+        try:
+            product_type: str = filter["productType"]
```
sbrunato:
```suggestion
            product_type = search_params.get("productType")
```

eodag/plugins/search/build_search_result.py
```diff
@@ -305,11 +305,11 @@ def append_time(input_date: date, time: Optional[str]) -> datetime:
     """
```
sbrunato, line 305: why did you update this method ? Is `HH_MM` format no more used ?

docs/notebooks/api_user_guide/3_search.ipynb
sbrunato, line 1: please rebase on develop branch

eodag/plugins/search/build_search_result.py
dalpasso: I added the support for `HH_MM` that was missing. Previously only the format `HH:MM` was supported. The provider `cop_cds` uses the format `HH_MM` as you can see from the [constrains.json](https://cds.climate.copernicus.eu/api/catalogue/v1/collections/sis-agrometeorological-indicators/constraints.json) file.

eodag/api/core.py
dalpasso: Enabling validation of a search by `id` would fail cause other required parameters would be missing. For example, a search by `id`for the for the provider `cop_cds` results in `ecmwf:statistic: Field required; ecmwf:variable: Field required; ecmwf:time: Field required; ecmwf:version: Field required`.

eodag/plugins/search/base.py
dalpasso: By using `search_params.get("productType")`, the function doesn't validate correctly cause missing product type configuration. I didn't find a way to get the product type configurations from within the `Search` plugin. In case this is not possible, alternatives solutions are:

- raising an exception for missing product type in `validate()`
- moving the function `validate()` to `EODataAccessGateway`

dalpasso: This case has been handled in the function `validate()`:
```python
            product_type = search_params.get("productType")
            if not product_type:
                raise ValidationError("Field required: productType")
```

PR #1783, opened 2025-08-22
fix(providers): product types discovered properties format

timothey-cs:


eodag/resources/providers.yml
```diff
@@ -2715,6 +2715,9 @@
     ssl_verify: true
     pagination:
       max_items_per_page: 1000
+    discover_product_types:
+      generic_product_type_parsable_metadata:
+        processingLevel: '{$.summaries."processing:level"#csv_list}'
```
sbrunato: better change this in stac_provider.yml

sbrunato (approved):

PR #1741, opened 2025-07-21
refactor(core): whoosh removal

sbrunato:
Fixes #1402, #546

Removes `whoosh` to simplify product types handling and removes the index that can cause locking errors when running concurrent commands

PR #1800, opened 2025-09-01
fix(core): guess_product_type using alias

jlahovnik:
`guess_product_type` returns alias again as it was before `whoosh` removal refactoring https://github.com/CS-SI/eodag/pull/1741


eodag/api/core.py
sbrunato, line 1: can you add a test for that please

jlahovnik: done

sbrunato (approved):

PR #1801, opened 2025-09-01
fix(providers): ssl verify for fedeo_ceda

timothey-cs:
Enable SSL verification on provider `fedeo_ceda` search now that their SSL certificate is fixed.


eodag/resources/providers.yml
```diff
@@ -5939,7 +5939,7 @@
   search: !plugin
     type: StacSearch
     api_endpoint: 'https://fedeo.ceos.org/search'
-    ssl_verify: false
+    ssl_verify: True
```
sbrunato:
```suggestion
    ssl_verify: true
```

sbrunato (changes requested):

PR #1805, opened 2025-09-01
refactor: use zipstream-ng instead of stream-zip

sbrunato:
Following https://github.com/CS-SI/eodag/discussions/1798,

[stream-zip](https://pypi.org/project/stream-zip) source code is no more available: https://github.com/uktrade/stream-zip

The author answer is:
> We're unable to provide more information due to Section 24 of FOIA.
> You can read more about Section 24 here: https://ico.org.uk/for-organisations/foi/freedom-of-information-and-environmental-information-regulations/section-24-safeguarding-national-security

This PR removes `stream-zip` dependency ant replaces it with [zipstream-ng](https://github.com/pR0Ps/zipstream-ng).

### DIFF
--- a/eodag/plugins/download/http.py
+++ b/eodag/plugins/download/http.py
@@ -812,7 +812,7 @@ class HTTPDownload(Download):
                     for asset_stream in assets_stream_list:
                         zip_stream.add(
                             asset_stream.content,
-                            arcname=asset_stream.filename,
+                            arcname=asset_stream.arcname,
                             size=asset_stream.size,
                         )
 
@@ -1108,8 +1108,10 @@ class HTTPDownload(Download):
             else None
         )
 
-        def get_chunks_generator(asset_href: str) -> Iterator[bytes]:
+        def get_chunks_generator(asset: Asset) -> Iterator[bytes]:
             """Create a generator function that will be called by ZipStream when needed."""
+
+            asset_href = asset.get("href")
             # This function will be called by zipstream when it needs the data
             if not asset_href or asset_href.startswith("file:"):
                 logger.info(f"Local asset detected. Download skipped for {asset_href}")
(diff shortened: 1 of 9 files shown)


eodag/utils/s3.py
```diff
@@ -298,6 +310,8 @@ def _build_stream_response(
         - "raw": Stream files directly, as a single file or multipart.
         - "auto": ZIP if multiple files, raw if single file.
     :param executor: Executor used for concurrent streaming and cleanup.
+    :param s3_client: S3 client (kept for signature compatibility).
+    :param range_size: Size of download chunks (kept for signature compatibility).
```
alambare: This is a private method, we do not need to keep the signature compatibility.

eodag/utils/s3.py
```diff
@@ -316,40 +330,32 @@ def _build_response(
         content_gen: Iterable[bytes],
         media_type: str,
         extra_headers: dict[str, str] = {},
```
alambare: it looks like this is not needed anymore

eodag/plugins/download/http.py
```diff
@@ -786,48 +785,46 @@ def _stream_download_dict(
...
+                    for asset_stream in assets_stream_list:
+                        zip_stream.add(
+                            asset_stream.content,
+                            arcname=asset_stream.arcname,
+                            size=asset_stream.size,
+                        )
+
+                    # do not use global size if one of the assets has no size
+                    missing_length = any(not (asset.size) for asset in assets_values)
+                    zip_length = len(zip_stream) if not missing_length else None
```
alambare: We should not set `sized=True` if we are not sure that all assets have a size.
Otherwise, according to `ZipStream` documentation, those assets will be fully read in memory.

PR #1830, opened 2025-09-16
fix(#1825): harmonize orbitDirection values to lower

### LINKED ISSUE #1825: harmonize orbitDirection values
Following a request in https://github.com/CS-SI/eodag/issues/1305#issuecomment-3270603153, harmonize `orbitDirection` to a lowercase value `ascending` in order to align with [SAT STAC extension](https://github.com/stac-extensions/sat?tab=readme-ov-file#satorbit_state).

timothey-cs:
Fixes #1825 
All orbitDirection are now in lower case for all the providers

### DIFF
--- a/eodag/resources/providers.yml
+++ b/eodag/resources/providers.yml
@@ -349,7 +349,7 @@
           - '$.absoluteOrbitNumber'
         orbitDirection:
           - '{{"search":{{"passDirection":"{orbitDirection}" }} }}'
-          - '$.passDirection'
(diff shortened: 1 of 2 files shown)


eodag/resources/providers.yml
sbrunato, line 1: update `convert_to_lower` to skip `NOT_AVAILABLE` values

sbrunato (changes requested):

PR #1890, opened 2025-10-21
perf: download products and assets in parallel

### LINKED ISSUE #146: Parallel processing for downloading and unziping
We can save a lot of processing time, specially for large dataset of product (time serie) if the downloading and the unziping (exemple of S1 products) can be done in 2 differents threads.

Actually, when downloading, the CPU is not used at full capacity, and when unzip, the network is not use.

Doing both in parallel can speedup a lot the processing codes.

anesson-cs:
Fixes #146 

Perform multiple products downloads from `download_all()` core method in parallel to increase significantly efficiency.
If a product has assets, they are also downloaded in parallel. In that case and if there are downloads in parallel, the same executor is used to not disturb the parallelism.

A user can decide to use his own executor by passing `executor` as positional argument of `download()` and `download_all()` core methods and `download()` method from an `EOProduct` instance. Otherwise, an executor by default without specification of `concurrent.futures.ThreadPoolExecutor` class will be used instead.

If a provider has limitations about the maximum number of workers from an executor to use, it will take over the current value of the executor if needed.

### DIFF
--- a/eodag/api/core.py
+++ b/eodag/api/core.py
@@ -1939,7 +1939,7 @@ class EODataAccessGateway:
                                   creation and update to give the user a
                                   feedback on the download progress
         :param executor: (optional) An executor to download EO products of ``search_result`` in parallel
-                                    which may be reused to also download assets of these products in parallel.
+                                    which will also be reused to download assets of these products in parallel.
         :param wait: (optional) If download fails, wait time in minutes between
                      two download tries of the same product
         :param timeout: (optional) If download fails, maximum time in minutes
@@ -2061,7 +2061,8 @@ class EODataAccessGateway:
                                   size as inputs and handle progress bar
                                   creation and update to give the user a
                                   feedback on the download progress
-        :param executor: (optional) An executor to download assets of ``product`` in parallel if it has any
+        :param executor: (optional) An executor to download assets of ``product`` in parallel if it has any. If ``None``
+                         , a default executor will be created
         :param wait: (optional) If download fails, wait time in minutes between
                     two download tries
         :param timeout: (optional) If download fails, maximum time in minutes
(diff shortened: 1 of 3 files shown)


eodag/api/core.py
```diff
@@ -2038,6 +2040,8 @@ def download_all(
                                   size as inputs and handle progress bar
                                   creation and update to give the user a
                                   feedback on the download progress
+        :param executor: (optional) An executor to download EO products of ``search_result`` in parallel
+                                    which may be reused to also download assets of these products in parallel.
```
sbrunato:
```suggestion
                                    which will also be reused to download assets of these products in parallel.
```

eodag/api/core.py
```diff
@@ -2158,6 +2164,7 @@ def download(
                                   size as inputs and handle progress bar
                                   creation and update to give the user a
                                   feedback on the download progress
+        :param executor: (optional) An executor to download assets of ``product`` in parallel if it has any
```
sbrunato:
```suggestion
        :param executor: (optional) An executor to download assets of ``product`` in parallel if it has any. If ``None``,
                         a default executor will be created
```

eodag/api/product/_product.py
```diff
@@ -350,6 +354,7 @@ def download(
                                   size as inputs and handle progress bar
                                   creation and update to give the user a
                                   feedback on the download progress
+        :param executor: (optional) An executor to download assets of the product in parallel if it has any
```
sbrunato:
```suggestion
        :param executor: (optional) An executor to download assets of the product in parallel if it has any. If ``None``,
                         a default executor will be created
```

eodag/cli.py
```diff
@@ -558,6 +559,11 @@ def discover_col(ctx: Context, **kwargs: Any) -> None:
     type=click.Path(dir_okay=True, file_okay=False),
     help="Products or quicklooks download directory (Default: local temporary directory)",
 )
+@click.option(
+    "--max-workers",
+    type=int,
+    help="The maximum number of workers to use for downloading products in parallel",
```
sbrunato:
```suggestion
    help="The maximum number of workers to use for downloading products and assets in parallel",
```

PR #1923, opened 2025-11-06
fix(plugins): cop_marine search by id

jlahovnik:
- improve performance of `search_by_id` by avoiding to continue search after item was found
- avoid duplicate `validate` parameter in iteration of search pages

### DIFF
--- a/eodag/plugins/search/cop_marine.py
+++ b/eodag/plugins/search/cop_marine.py
@@ -388,11 +388,11 @@ class CopMarineSearch(StaticStacSearch):
                         dataset_item,
                         collection_dict,
                         True,
-                        kwargs.get("id", None),
+                        kwargs.get("id"),
                     )
                     if product:
                         products.append(product)
-                    if product and kwargs.get("id", None):
+                    if product and kwargs.get("id"):
                         break
                     continue
 


eodag/api/search_result.py
```diff
@@ -301,6 +301,9 @@ def get_next_page(current):
             if current.number_matched:
                 search_kwargs["number_matched"] = current.number_matched
             for i, search_plugin in enumerate(search_plugins):
+                search_kwargs.pop(
```
sbrunato: is it needed ? `_do_search` is executed with `validate=False` just below

eodag/plugins/search/cop_marine.py
```diff
@@ -385,9 +388,12 @@ def query(
                         dataset_item,
                         collection_dict,
                         True,
+                        kwargs.get("id", None),
```
sbrunato:
```suggestion
                        kwargs.get("id"),
```

eodag/plugins/search/cop_marine.py
```diff
@@ -385,9 +388,12 @@ def query(
...
                     )
                     if product:
                         products.append(product)
+                    if product and kwargs.get("id", None):
```
sbrunato:
```suggestion
                    if product and kwargs.get("id"):
```

sbrunato (changes requested):

sbrunato (approved):

PR #1924, opened 2025-11-07
fix(plugins): ecmwf geometries support

dalpasso:
This PR converts some geometry data structures in a valid format:

- ECMWF Polytope uses non-geojson structure for features. Converts it to a Shapely polygon.
- Bounding box in area format (`[max_lat,min_lon,min_lat,max_lon]`) is converted to Shapely polygon.

### DIFF
--- a/eodag/utils/__init__.py
+++ b/eodag/utils/__init__.py
@@ -84,7 +84,7 @@ from shapely.geometry.base import GEOMETRY_TYPES, BaseGeometry
 from tqdm.auto import tqdm
 
 from eodag.utils import logging as eodag_logging
-from eodag.utils.exceptions import MisconfiguredError, ValidationError
+from eodag.utils.exceptions import MisconfiguredError
 
 if TYPE_CHECKING:
     from jsonpath_ng import JSONPath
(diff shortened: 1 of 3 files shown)


eodag/utils/__init__.py
```diff
@@ -1154,6 +1154,37 @@ def get_geometry_from_various(
     return geom
 
 
+def get_geometry_from_shape(geom: dict[str, Any]) -> BaseGeometry:
```
sbrunato:
```suggestion
def get_geometry_from_ecmwf_feature(geom: dict[str, Any]) -> BaseGeometry:
```

eodag/utils/__init__.py
```diff
@@ -1154,6 +1154,37 @@ def get_geometry_from_various(
...
+        shape: list = geom["shape"]
+        polygon_args = [(p[1], p[0]) for p in shape]
+        return Polygon(polygon_args)
+    else:
+        raise ValidationError(
+            "convert_from_geojson_polytope only accepts shapes of type polygon"
+        )
+
+
+def get_geometry_from_area(area: list[float]) -> Optional[BaseGeometry]:
```
sbrunato:
```suggestion
def get_geometry_from_ecmwf_area(area: list[float]) -> Optional[BaseGeometry]:
```

eodag/utils/__init__.py
sbrunato, line 1: please check that `get_geometry_from_ecmwf_*` inputs are well formatted (type, shape) to avoid unexpected errors when validating:
```
"ecmwf:feature": "foo"
```
is not catched for the moment

PR #1926, opened 2025-11-17
fix(providers): config for DT_CLIMATE_G1_CMIP6_HIST_IFS_FESOM_R1 collection

jlahovnik:
for collection DT_CLIMATE_G1_CMIP6_HIST_IFS_FESOM_R1 the dataset parameter should not be added to the request


eodag/resources/providers.yml
sbrunato, line 1: rebase needed

jlahovnik: done

PR #1947, opened 2025-11-28
fix(core): add Collection property for server extended metadata

jlahovnik:
adds a property `eodag_stac_collection` to the `Collection` object which is used by `stac-fastapi-eodag` to import collection metadata from an external file

### DIFF
--- a/eodag/api/collection.py
+++ b/eodag/api/collection.py
@@ -91,7 +91,7 @@ class Collection(BaseModel):
         repr=False,
     )
 
-    # path to external collection metadata file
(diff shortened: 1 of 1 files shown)


eodag/api/collection.py
```diff
@@ -91,6 +91,11 @@ class Collection(BaseModel):
         repr=False,
     )
 
+    # path to external collection metadata file
```
sbrunato:
```suggestion
    # path to external collection metadata file (required by stac-fastapi-eodag)
```

PR #1952, opened 2025-12-02
fix(plugins): delete download link when order not finish

cauriol:
Remove download link if the order does not succeed when using `HTTPDownload`


eodag/plugins/download/http.py
```diff
@@ -476,6 +476,8 @@ def _request(
         if (
             success_status and success_status != status_dict.get("eodag:order_status")
         ) or (success_code and success_code != response.status_code):
+            if "eodag:download_link" in product.properties:
```
jlahovnik: you can use product.properties.pop("eodag:download_link", None) instead of if and del
Also: please add a short comment why this is done

cauriol: Done

PR #1956, opened 2025-12-03
fix(plugins): dates formatting and missing param in ECMWFSearch queryables

dalpasso:
Fix in the queryables:

- add `satellite_mission` as allowed Copernicus keyword;
- interpret `/` and `/to/` as range separator for dates (e.g. `2004-01-01/2004-01-10`). Otherwise the values are used as simple strings (e.g. `"0.5/0.5"` is not interpreted as range, but as simple string).

### DIFF
--- a/eodag/utils/dates.py
+++ b/eodag/utils/dates.py
@@ -35,7 +35,17 @@ RFC3339_PATTERN = (
     r"(Z|([+-])(\d{2}):(\d{2}))?)?$"
 )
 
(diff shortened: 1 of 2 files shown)


eodag/utils/dates.py
```diff
@@ -35,6 +35,8 @@
     r"(Z|([+-])(\d{2}):(\d{2}))?)?$"
 )
 
+DATE_RANGE_PATTERN = r"\d{4}-\d{2}-\d{2}((/to/|/)\d{4}-\d{2}-\d{2})?"
```
jlahovnik: The date pattern shoud also work for dedt_lumi format (date without "-").

PR #1968, opened 2025-12-09
refactor(core): use collection alias in EOProduct

jlahovnik:
- use the alias for the `collection` property of `EOProduct`
- remove the `eodag:alias` property

### DIFF
--- a/eodag/api/core.py
+++ b/eodag/api/core.py
@@ -1849,12 +1849,6 @@ class EODataAccessGateway:
                     else:
                         eo_product.collection = guesses[0].id
 
-                try:
-                    if eo_product.collection is not None:
-                        self.get_collection_from_alias(eo_product.collection)
-                except NoMatchingCollection:
-                    logger.debug("collection %s not found", eo_product.collection)
-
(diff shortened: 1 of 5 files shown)


eodag/api/core.py
```diff
@@ -1851,9 +1851,7 @@ def _do_search(
 
                 try:
                     if eo_product.collection is not None:
-                        eo_product.collection = self.get_collection_from_alias(
-                            eo_product.collection
-                        )
+                        self.get_collection_from_alias(eo_product.collection)
```
sbrunato: what is it for ? only check for duplicate ids in conf ?

eodag/api/product/_product.py
sbrunato, line 1: please remove `eodag:alias` occurrences in code, as it should be `collection`

jlahovnik: refactoring to remove eodag:alias property done

eodag/api/core.py
jlahovnik: I removed the condition because checks are already done before.

PR #1975, opened 2025-12-11
fix(plugins): aws streamed downloads

anesson-cs:
Downloading products in streaming with `AwsDownload` plugin did not work anymore. This PR allows to fix it by coding an other way of checking if an object is a `boto3.resource` in the `_stream_download_dict()` method.

### DIFF
--- a/eodag/plugins/download/aws.py
+++ b/eodag/plugins/download/aws.py
@@ -773,7 +773,6 @@ class AwsDownload(Download):
         )
 
         # check if auth is a S3 resource by verifying it has the meta.client attribute.
-        # isinstance() check is not possible here because it is a dynamically generated class.
         if auth and hasattr(auth, "meta") and hasattr(auth.meta, "client"):
             s3_resource = auth
         else:


sbrunato (changes requested):
> Downloading products with AwsDownload plugin did not work anymore

Please describe the problem more explicitly, as `AwsDownload.download` still seems to work.

Also, there is no server mode any more, so the PR should not reference it

eodag/plugins/download/aws.py
```diff
@@ -771,7 +771,10 @@ def _stream_download_dict(
             ignore_assets,
             product,
         )
-        if auth and isinstance(auth, boto3.resource("s3").__class__):
+
+        # check if auth is a S3 resource by verifying it has the meta.client attribute.
+        # isinstance() check is not possible here because it is a dynamically generated class.
```
sbrunato:
```suggestion
```

sbrunato (changes requested):

sbrunato (approved):

PR #1983, opened 2026-01-05
fix(core): asset name harmonization with query-string

cauriol:
This PR fixes asset naming issues for Sentinel-1 assets coming from Planetary Computer and creodias_s3.

Asset names are now normalized to the name.extension format.
For example:

`preview.png?foo` → `preview.png`

`calibration-vh-12a34b.xml` → `calibration-vh.xml`

This PR updates the `ASSET_KEYS_PATTERNS_ROLES` and `REPLACE_PATTERNS` lists in drivers and adds new tests.

### DIFF
--- a/tests/units/test_eoproduct_driver_sentinel1.py
+++ b/tests/units/test_eoproduct_driver_sentinel1.py
@@ -52,7 +52,7 @@ class TestEOProductDriverSentinel1Driver(EODagTestCase):
         )
         self.assertEqual(
             self.product.driver.guess_asset_key_and_roles(
-                "http://foo/1/28/0/iw-hh.tif", self.product
+                "http://foo/1/28/0/iw-hh.tif?foo", self.product
             ),
             ("HH", ["data"]),
         )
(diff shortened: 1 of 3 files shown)


jlahovnik (changes requested):
Please add a test case and update the PR description!

jlahovnik (approved):

eodag/api/product/drivers/sentinel1.py
sbrunato, line 1: on `S1_SAR_GRD` from `earth_search`, `manifest.safe` gets renamed to `manifest.s` when it shouldn't

tests/units/test_eoproduct_driver_sentinel1.py
```diff
@@ -74,6 +86,12 @@ def test_driver_s1_guess_asset_key_and_roles(self):
             ),
             ("thumbnail.png", ["thumbnail"]),
         )
+        self.assertEqual(
+            self.product.driver.guess_asset_key_and_roles(
+                "s3://foo/1/28/0/preview.png?collection=foo&items=foo", self.product
+            ),
+            ("previ.png", ["overview"]),
```
sbrunato: `previ.png` is not the expected output. Please update driver patterns of chars that could be included in other words (`ew`, `iw`, `sm`, `raw`) with something like `r"(?<![A-Za-z])ew(?![A-Za-z])"`, to avoid letters on both sides

eodag/api/product/drivers/sentinel1.py
sbrunato, line 1: these fixes should also be applied to other drivers

cauriol: I applied the changes to the sentinel2 et generic

tests/units/test_eoproduct_driver_sentinel1.py
cauriol: Done
We have preview.png now

eodag/api/product/drivers/sentinel1.py
cauriol: Fix

sbrunato (approved):

PR #1985, opened 2026-01-08
fix(providers): update wekeo config for CLMS global collections

jlahovnik:
The configuration on `wekeo` for the CLMS global collections has changed: Instead of having one dataset per collection, there is one dataset for all collections which can be filtered by `productType` and `productIdentifier` to get the data for a specific collection. 
Also fixes an bug with the properties extraction in case a collection alias is used.

### DIFF
--- a/eodag/plugins/search/qssearch.py
+++ b/eodag/plugins/search/qssearch.py
@@ -1275,16 +1275,17 @@ class QueryStringSearch(Search):
         )
         products: list[EOProduct] = []
         asset_key_from_href = getattr(self.config, "asset_key_from_href", True)
+        product_kwargs = deepcopy(kwargs)
+        # collection alias as collection property for product
+        if alias := getattr(self.config, "collection_config", {}).get("alias"):
+            product_kwargs["collection"] = alias
         for result in results:
             properties = QueryStringSearch.extract_properties[self.config.result_type](
                 result,
                 self.get_metadata_mapping(kwargs.get("collection")),
                 discovery_config=getattr(self.config, "discover_metadata", {}),
             )
-            # collection alias (required by opentelemetry-instrumentation-eodag)
-            if alias := getattr(self.config, "collection_config", {}).get("alias"):
-                kwargs["collection"] = alias
-            product = EOProduct(self.provider, properties, **kwargs)
+            product = EOProduct(self.provider, properties, **product_kwargs)
 
             additional_assets = self.get_assets_from_mapping(result)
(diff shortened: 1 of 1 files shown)


jlahovnik:
currently there is an issue with collection CLMS_GLO_DMP_333M - the search result contains incorrect items. It seems to be an issue on provider side -> waiting for feedback

eodag/resources/providers.yml
```diff
@@ -3375,17 +3375,22 @@
       product:type: DTE_90
       metadata_mapping_from_product: COP_DEM_GLO30_DGED
     CLMS_GLO_NDVI_333M:
-      _collection: EO:CLMS:DAT:CLMS_GLOBAL_NDVI_300M_V1_10DAILY_NETCDF
-      product:type: NDVI300
+      _collection: EO:JRC:DAT:CLMS
+      product:type: vegetation_indices
+      product:identifier: ndvi_global_300m_10daily_v1
```
sbrunato: `product:identifier` does not exist https://github.com/stac-extensions/product?tab=readme-ov-file#fields
please use another name like `_product_identifier`

eodag/resources/providers.yml
```diff
@@ -3375,17 +3375,22 @@
...
       metadata_mapping:
-        <<: *id_from_date
-        eodag:order_link: 'https://gateway.prod.wekeo2.eu/hda-broker/api/v1/dataaccess/download?{{"location": "{eodag:download_link}","product_id":"{id}", "dataset_id": "EO:CLMS:DAT:CLMS_GLOBAL_NDVI_300M_V1_10DAILY_NETCDF"}}'
+        product:identifier:
+        - '{{"productIdentifier": "{product:identifier}"}}'
+        - '$.null'
+        id:
+        - '{{"productIdentifier": "{id}"}}'
+        - '$.id'
```
sbrunato: what's the point of having this conf for `id` ?
When the collection is also set, `productIdentifier` will come from its conf. And when it is not set, search will not be performed because `dataset_id` is missing

eodag/resources/providers.yml
jlahovnik: changed

eodag/resources/providers.yml
jlahovnik: For searching by id but I realised that it does not work because it is overwritten witht the default value. I also noticed a strange behaviour of the date filter on wekeo side -> waiting for feedback on that.

sbrunato (approved):

PR #1993, opened 2026-01-13
refactor: cli.py collision name on eodag

pdavid-cssopra:
__________________________________________________________________________

Resolve collision on name "eodag", is module and cli root object "eodag"
Proposal: remane cli root function as "eodag_cli":

def eodag_cli(ctx: Context, verbose: int) -> None:
[...]
@eodag_cli.command
[...]
if __name__ == "__main__":
    eodag_cli(obj={})

__________________________________________________________________________

By fixing tests, see some lack on syspath / bootstrap set, it fixed.
Steadify cli test by ensure assert on each command line
__________________________________________________________________________

Fix eoproduct test to be able to validate PR
test_eoproduct.py:482
```
        # progress bar finished
        self.assertEqual(progress_callback.n, progress_callback.total)
```
Progressbar has a temporal behaviour, then depens of 100ms passed, progress_callback.n can be ==0 or ==1
By waiting de safe duration to ensure progress bar completed, test is safe now

```
        # Progressbar need at least "progress_callback.mininterval" seconds, here 0.1 second
        # Wait 0.2 to be sure progress ends
        time.sleep(0.2)

        # progress bar finished
        self.assertEqual(progress_callback.n, 1)
```
__________________________________________________________________________

Test concurency troubleshot: 
```
  _______________ TestCollectionsList.test_search_result_repr_html _______________
  [gw3] linux -- Python 3.14.2 /home/runner/work/eodag/eodag/.tox/py314/bin/python3
  
  self = <tests.units.test_collection.TestCollectionsList testMethod=test_search_result_repr_html>
  
      def setUp(self):
          super(TestCollectionsList, self).setUp()
  >       self.dag = EODataAccessGateway()
                     ^^^^^^^^^^^^^^^^^^^^^
  
  tests/units/test_collection.py:298: 
  _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
  eodag/api/core.py:158: in __init__
      copyfile(src, dst, follow_symlinks=follow_symlinks)
  _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
  
  src = '/home/runner/work/eodag/eodag/eodag/resources/user_conf_template.yml'
  dst = '/home/runner/work/eodag/eodag/.tox/py314/tmp/.config/eodag/eodag.yml'
  
      def copyfile(src, dst, *, follow_symlinks=True):
          """Copy data from src to dst in the most efficient way possible.
      
          If follow_symlinks is not set and src is a symbolic link, a new
          symlink will be created instead of copying the file it points to.
      
          """
          sys.audit("shutil.copyfile", src, dst)
      
          if _samefile(src, dst):
              raise SameFileError("{!r} and {!r} are the same file".format(src, dst))
      
          file_size = 0
          for i, fn in enumerate([src, dst]):
              try:
                  st = _stat(fn)
              except OSError:
                  # File most likely does not exist
                  pass
              else:
                  # XXX What about other special files? (sockets, devices...)
                  if stat.S_ISFIFO(st.st_mode):
                      fn = fn.path if isinstance(fn, os.DirEntry) else fn
                      raise SpecialFileError("`%s` is a named pipe" % fn)
                  if _WINDOWS and i == 0:
                      file_size = st.st_size
      
          if not follow_symlinks and _islink(src):
              os.symlink(os.readlink(src), dst)
          else:
              with open(src, 'rb') as fsrc:
                  try:
  >                   with open(dst, 'wb') as fdst:
                           ^^^^^^^^^^^^^^^
  E                   FileNotFoundError: [Errno 2] No such file or directory: '/home/runner/work/eodag/eodag/.tox/py314/tmp/.config/eodag/eodag.yml'
  
  ../../../.local/share/uv/python/cpython-3.14.2-linux-x86_64-gnu/lib/python3.14/shutil.py:315: FileNotFoundError
```
Config files are manipulated from multiple instance of EODataAccessGateway, from multiple test thread, have to steadify concurency manage.

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -15,7 +15,6 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
 import logging
 import os
 import re
@@ -24,8 +23,10 @@ from contextlib import contextmanager
 from datetime import datetime
 from importlib.resources import files as res_files
 from tempfile import TemporaryDirectory
+from typing import Optional, Tuple
 
 import click
+import shapely
 from click.testing import CliRunner
 from faker import Faker
 from packaging import version
@@ -33,6 +34,7 @@ from packaging import version
 from eodag.api.collection import Collection, CollectionsList
 from eodag.api.search_result import SearchResult
 from eodag.utils import GENERIC_COLLECTION
+from eodag.utils.exceptions import UnsupportedProvider
 from tests import TEST_RESOURCES_PATH
 from tests.context import (
     DEFAULT_ITEMS_PER_PAGE,
@@ -40,7 +42,7 @@ from tests.context import (
     MisconfiguredError,
     NoMatchingCollection,
     download,
-    eodag,
+    eodag_cli,
     mock,
     search_crunch,
 )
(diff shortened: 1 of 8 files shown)


tests/test_cli.py
```diff
@@ -15,32 +15,38 @@
...
+from typing import Optional, Tuple
 
 import click
+import shapely
 from click.testing import CliRunner
 from faker import Faker
 from packaging import version
 
+if __name__ == "__main__":
+    sys.path.insert(0, os.path.normpath(os.path.join(os.path.dirname(__file__), "..")))
```
sbrunato: i'd prefer to exclude these lines from every test file
- tests can still be run using pytest or tox
- it's not really readable
- and will certainly be forgotten in some some new tests files in the future

tests/test_cli.py
```diff
@@ -90,34 +96,64 @@ def tearDown(self):
...
         logger.level = 0
 
+    def eodag_command(
+        self, command_list: list = [], env: dict = {}
+    ) -> Tuple[int, str, Optional[Exception]]:
+        """Eodag command line
+        :param list          command line
+        :result int          system exit_code
+        :result str          stdout
+        :result str          stderr
```
sbrunato:
```suggestion
        """Eodag command line

        :param command_list: command line
        :param env: environment variables
        :returns: exit_code, stdout, stderr
```

tests/test_cli.py
```diff
@@ -1031,8 +1154,15 @@ def test_eodag_download_quicklooks(self, mock_get_quicklook):
...
             "A quicklook may have been downloaded but we cannot locate it. "
             "Increase verbosity for more details: `eodag -v download [OPTIONS]`",
-            result.output,
+            output,
         )
+
+
+if __name__ == "__main__":
+    # Allow direct run test by run file
+    unittest.main()
```
sbrunato: please also remove this footer in all files. Test should be run using pytest or tox

sbrunato (changes requested):

tests/test_cli.py
pdavid-cssopra: This correction is not insignificant. It's designed to correctly implement the sys.path definition initially defined in tests/context.py:27:
```
if __name__ == "__main__":
sys.path.insert(0, os.path.normpath(os.path.join(os.path.dirname(__file__), "..")))
``` 

This is only correct for tests within "tests", but incorrect for tests in the subfolders "tests/integration" and "tests/units".

It's precisely the use of tox/pytest that masks the sys.path definition error by first running a test in the root folder. This problem becomes apparent when pytest is instructed to run only the tests in a subfolder, revealing that it cannot find the eodag packages on an incorrect relative path.

Unfortunately, as long as the tests are outside the package being tested, providing the target's sys.path is not optional. Python standards have not yet offered a more aesthetically pleasing way to ensure dependency relativity other than by explicitly defining them.

tests/test_cli.py
pdavid-cssopra: This standardization stems from an observation made in "tests/test_provider.py", which reveals a bootstrap corresponding to the basic test case used.

However, there is clearly some confusion between the two.

Unittest requires developers to create classes derived from the TestCase module, and then define test cases as methods within the class. Pytest, on the other hand, only requires you to define a function preceded by "test_" and to use the assertion conditions within it.

The pytest formalism (already present before modification)
```
# tests/test_provider.py:22
import pytest

[...]

# tests/test_provider.py:42
class TestProvider:
    """Test cases for the Provider class."""

    @pytest.fixture # <-- imply direct pytest.fixture inherits method
    def basic_config(self):

# tests/test_provider.py:384
if __name__ == "__main__":
    # Allow direct run test by run file
    pytest.main([__file__]) # <-- imply explicit bootstrap definition from used test class

```

The unitest formalist:
```
import unittest

[...]

class BaseDownloadPluginTest(unittest.TestCase): # <-- imply direct unittest.TestCase inherits class

[...]

if __name__ == "__main__":
    # Allow direct run test by run file
    unittest.main() # <-- adapt explicit bootstrap definition from pytest files
```

Or the bootstrap explicitly indicates the entry point of the test class used.


The vast majority of tests are written using the derived class `unittest`, and employ assertions associated with this class rather than the `assert` command. Therefore, these are not `pytest` tests, but rather `unittest` tests; here, it's the `pytest` application that makes the difference, not the code itself.

In this case, it might be a good idea to clearly define the scope of the tests.

tests/test_cli.py
sbrunato: > It's precisely the use of tox/pytest that masks the sys.path definition error by first running a test in the root folder. This problem becomes apparent when pytest is instructed to run only the tests in a subfolder, revealing that it cannot find the eodag packages on an incorrect relative path.

to run tests, you have to install `eodag[dev]` (or `.[dev]` from your workspace). Then, pytest has no problem to find eodag package

tests/test_cli.py
sbrunato: The original footer from `tests/test_provider.py` can also be removed.
This file was written out of the other tests scope, it is not a good practice for harmonization between tests, but it works as is, and we decided to not rewrite it.

sbrunato (approved):

PR #2001, opened 2026-01-19
fix(crunch): minor fixes and improved crunches test coverage

### LINKED ISSUE #2000: raise coverage plugin crunch

pdavid-cssopra:
Fixes #2000 
Minor fixes and improved crunches test coverage

### DIFF
--- a/eodag/plugins/base.py
+++ b/eodag/plugins/base.py
@@ -66,9 +66,12 @@ class PluginTopic(metaclass=EODAGPluginMount):
 
     def __repr__(self) -> str:
         config = getattr(self, "config", None)
+        priority = ""
+        if config is not None and hasattr(config, "priority"):
+            priority = str(config.priority)  # is an int
         return "{}(provider={}, priority={}, topic={})".format(
             self.__class__.__name__,
             getattr(self, "provider", ""),
-            config.priority if config else "",
+            priority,
             self.__class__.mro()[-3].__name__,
         )
--- a/eodag/plugins/crunch/base.py
+++ b/eodag/plugins/crunch/base.py
@@ -17,6 +17,7 @@
 # limitations under the License
 from __future__ import annotations
 
+from abc import abstractmethod
 from typing import TYPE_CHECKING, Any, Optional
 
 from eodag.config import PluginConfig
@@ -32,10 +33,11 @@ class Crunch(PluginTopic):
     :param config: Crunch configuration
     """
 
-    def __init__(self, config: Optional[dict[str, Any]]) -> None:
+    def __init__(self, config: Optional[dict[str, Any]] = None) -> None:
         self.config = PluginConfig()
         self.config.__dict__ = config if config is not None else {}
 
+    @abstractmethod
     def proceed(
         self, products: list[EOProduct], **search_params: Any
     ) -> list[EOProduct]:
--- a/eodag/plugins/crunch/filter_latest_intersect.py
+++ b/eodag/plugins/crunch/filter_latest_intersect.py
@@ -66,7 +66,7 @@ class FilterLatestIntersect(Crunch):
         :returns: The filtered products
         """
         logger.debug("Start filtering for latest products")
-        if not products:
+        if not products or (isinstance(products, list) and len(products) == 0):
             return []
         # Warning: May crash if start_datetime is not in the appropriate format
         products.sort(key=self.sort_product_by_start_date, reverse=True)
--- a/eodag/plugins/crunch/filter_latest_tpl_name.py
+++ b/eodag/plugins/crunch/filter_latest_tpl_name.py
@@ -42,8 +42,12 @@ class FilterLatestByName(Crunch):
 
     NAME_PATTERN_CONSTRAINT = re.compile(r"\(\?P<tileid>\\d\{6\}\)")
 
-    def __init__(self, config: dict[str, Any]) -> None:
+    def __init__(self, config: Optional[dict[str, Any]] = None) -> None:
+        if config is None:
+            config = {}
         super(FilterLatestByName, self).__init__(config)
+        if "name_pattern" not in config:
+            raise ValidationError('Required parameter "name_pattern" in config')
         name_pattern = config.pop("name_pattern")
         if not self.NAME_PATTERN_CONSTRAINT.search(name_pattern):
             raise ValidationError(
--- a/eodag/plugins/crunch/filter_overlap.py
+++ b/eodag/plugins/crunch/filter_overlap.py
@@ -63,7 +63,13 @@ class FilterOverlap(Crunch):
                 "geometry not found in cruncher arguments, filtering disabled."
             )
             return products
-        minimum_overlap = float(self.config.__dict__.get("minimum_overlap", "0"))
+        try:
+            minimum_overlap = max(
+                float(self.config.__dict__.get("minimum_overlap", "0")), 0
+            )
+        except ValueError:
+            minimum_overlap = 0
+
         contains = self.config.__dict__.get("contains", False)
         intersects = self.config.__dict__.get("intersects", False)
         within = self.config.__dict__.get("within", False)
@@ -85,7 +91,7 @@ class FilterOverlap(Crunch):
             logger.debug("Minimum overlap is: {} %".format(minimum_overlap))
 
         logger.debug("Initial requested extent area: %s", search_geom.area)
-        if search_geom.area == 0:
+        if search_geom.area == 0.0:
             logger.debug(
                 "No product can overlap a requested extent that is not a polygon (i.e with area=0)"
             )
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -23,7 +23,7 @@ from importlib.resources import files as res_files
 from io import StringIO
 from tempfile import TemporaryDirectory
 
-import yaml.parser
+import yaml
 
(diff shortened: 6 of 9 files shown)


eodag/plugins/base.py
```diff
@@ -66,9 +66,16 @@ def __init__(self, provider: str, config: PluginConfig) -> None:
 
     def __repr__(self) -> str:
         config = getattr(self, "config", None)
+        priority = ""
+        if (
+            config is not None
+            and hasattr(config, "priority")
+            and isinstance(config.priority, str)
+        ):
+            priority = config.priority
```
sbrunato:
```suggestion
        ):
            priority = getattr(config, "priority", "")
```
priority has `int` type

eodag/plugins/crunch/base.py
```diff
@@ -32,10 +33,11 @@ class Crunch(PluginTopic):
     :param config: Crunch configuration
     """
 
-    def __init__(self, config: Optional[dict[str, Any]]) -> None:
+    def __init__(self, config: Optional[dict[str, Any]] = {}) -> None:
```
sbrunato: where was the default value expected ?

Anyway, this is not a good practice, see https://docs.astral.sh/ruff/rules/mutable-argument-default/

eodag/plugins/crunch/filter_latest_intersect.py
```diff
@@ -65,7 +65,7 @@ def proceed(
         :returns: The filtered products
         """
         logger.debug("Start filtering for latest products")
-        if not products:
+        if not products or (isinstance(products, list) and len(products) == 0):
```
sbrunato, line 69: not needed:
```py
>>> not []
True
```

eodag/plugins/crunch/filter_latest_tpl_name.py
```diff
@@ -42,8 +42,10 @@ class FilterLatestByName(Crunch):
 
     NAME_PATTERN_CONSTRAINT = re.compile(r"\(\?P<tileid>\\d\{6\}\)")
 
-    def __init__(self, config: dict[str, Any]) -> None:
+    def __init__(self, config: dict[str, Any] = {}) -> None:
         super(FilterLatestByName, self).__init__(config)
+        if "name_pattern" not in config:
+            raise ValidationError('require parameter "name_pattern"')
```
sbrunato:
```suggestion
            raise ValidationError('Required parameter "name_pattern" in config')
```

eodag/plugins/crunch/filter_overlap.py
```diff
@@ -63,7 +63,18 @@ def proceed(
...
+            minimum_overlap = max(
+                float(self.config.__dict__.get("minimum_overlap", "0")), 0
+            )
+        except ValueError:
+            minimum_overlap = 0
+
+        if minimum_overlap == 0:
+            logger.warning(
+                'require parameter "minimum_overlap" as a number in ]0, 100]'
+            )
```
sbrunato:
```suggestion
```
user should be able to run the crunch without specifying `minimum_overlap` (0 by default), and should not get a warning

tests/test_config.py
sbrunato, line 1: please discard these changes here and in other test files, as already discussed in
https://github.com/CS-SI/eodag/pull/1993#discussion_r2742416175 and https://github.com/CS-SI/eodag/pull/1993#discussion_r2742526285

tests/units/test_crunch.py
```diff
@@ -0,0 +1,692 @@
+# -*- coding: utf-8 -*-
+# Copyright 2018, CS GROUP - France, http://www.c-s.fr
```
sbrunato:
```suggestion
# Copyright 2026, CS GROUP - France, http://www.c-s.fr
```

eodag/plugins/crunch/filter_latest_intersect.py
pdavid-cssopra: Should try run tox with python 3.14, "if not <var>" not respect explicit typing require from 3.13 recommandation, because it's mixed type check as None/empty string/empty iterable/boolean false.
This is not an acceptable practice when applying explicit typing.

tests/units/test_crunch.py
```diff
@@ -0,0 +1,692 @@
...
+        )
+        # @TODO fix intersect, here result should be zero, it's not (1)
+        print(len(filtered_result))
+
+        filtered_result = search_results.crunch(
+            FilterLatestIntersect(),
+            geometry=Polygon(((0, 0), (0, 1), (1, 1), (1, 0), (0, 0))),
+        )
+        # @TODO fix intersect, here result should be zero, it's not (1)
+        print(len(filtered_result))
```
sbrunato: crunch was fixed in https://github.com/CS-SI/eodag/pull/2030, you can update the test

eodag/plugins/crunch/base.py
pdavid-cssopra: Mb, Optional does not disable positional argument required. LL replace it by None.

eodag/plugins/base.py
pdavid-cssopra: if priority is an int, getattr(config, "priority", "") makes mixting typing, using string fallback for a int value.
ll fix typing check.

tests/test_config.py
pdavid-cssopra: mb, file missed. patched

sbrunato: please also discard these changes in other test files

eodag/api/core.py
sbrunato, line 1: that's not the point of this PR. This may be suggested and discussed in a separate feature request

eodag/plugins/base.py
sbrunato: > if priority is an int

it is an `int`, see https://github.com/CS-SI/eodag/blob/v4.0.0b1/eodag/config.py#L343
You do not need to check its type again

pdavid-cssopra: it's specific command #getattr(config, "priority", "")# that makes mixin, at least concord type like #getattr(config, "priority", 0)#

eodag/api/core.py
pdavid-cssopra: It still makes troubles with CI execution and force invalid current branch.
Main trouble is concurrency during copying config file at eodag instanciation, no fix found to protect concurrency on it. Fix all tests one by one ll not prevent futures troubles. I remove concurrency fix, but still need to receive a real patch.

sbrunato: If you still find concurrency issues in CI, please point me to the failing jobs

tests/units/test_crunch.py
```diff
@@ -0,0 +1,689 @@
...
+        filtered_result = search_results.crunch(FilterProperty({"myproperty": 2}))
+        self.assertEqual(len(filtered_result), 0)
+
+        filtered_result = search_results.crunch(FilterProperty({"myproperty": "2"}))
+        self.assertEqual(len(filtered_result), 1)
+
+        filtered_result = search_results.crunch(FilterProperty({"myproperty": True}))
+        # @TODO should be ==0, here ==1, type are not checked before values True == bool
+        # Do not compare vars is not type concording...
+        print(">>", len(filtered_result))
```
sbrunato: that is not an issue, that is how [operator.eq()](https://docs.python.org/3/library/operator.html#module-operator) works. I just updated this [crunch documentation](https://eodag.readthedocs.io/en/latest/plugins_reference/generated/eodag.plugins.crunch.filter_property.FilterProperty.html) to explain its behavior

tests/units/test_crunch.py
```diff
@@ -0,0 +1,689 @@
...
+
+        filtered_result = search_results.crunch(FilterProperty({"myproperty": None}))
+        # @TODO should be ==1, here ==0, some comparition require "is" depending of type
+        # a product with empty property should not being consider has it not has property
+        # else, how request product with empty property, to be able to fix it ?
+        print(">>", len(filtered_result))
```
sbrunato: product.properties should not have `None` properties, they are already filtered out. See https://github.com/CS-SI/eodag/blob/v4.0.0b1/eodag/api/product/_product.py#L152

And also related to [operator.eq()](https://docs.python.org/3/library/operator.html#module-operator) behavior.

tests/units/test_crunch.py
pdavid-cssopra: So, was surprised to not have any warning by applying mismatch type value filter.

Test updated.

tests/units/test_crunch.py
pdavid-cssopra: So, was surprised to not have any warning by applying invalid value filter. Contrary to the previous remark, None applies as a value because no results are returned, whereas with an invalid value type in a filter, the results are returned without applying any filtering, the overall behavior lacks consistency.

Test updated.

eodag/plugins/crunch/filter_latest_tpl_name.py
```diff
@@ -42,8 +42,10 @@ class FilterLatestByName(Crunch):
 
     NAME_PATTERN_CONSTRAINT = re.compile(r"\(\?P<tileid>\\d\{6\}\)")
 
-    def __init__(self, config: dict[str, Any]) -> None:
+    def __init__(self, config: dict[str, Any] = {}) -> None:
```
sbrunato: same remark as https://github.com/CS-SI/eodag/pull/2001#discussion_r2768167704 for mutable as default

PR #1951, opened 2025-12-02
fix(plugins): uncomplete search validation for `cop_ads` and `wekeo_ecmwf`

dalpasso:
### Description

For some `cop_ads` collections (e.g. `CAMS_EU_AIR_QUALITY_FORECAST`, `CAMS_GAC_FORECAST`, `CAMS_GFE_GFAS`, `CAMS_SOLAR_RADIATION`) the HTTP search/order request to the remote provider (`cop_ads` and `wekeo_ecmwf`) fails even if the request validation succeed. This PR fixes the request validation so that it raises a `ValidationError` in such cases.

### Further comments

The parameters that caused the misbehave were mainly dates, geometry, and the requirement of the parameters itself.

Example: in the current implementation the following search passes the request validation, instead it should raise a `ValidationError` cause `start` and `end` parameters are missing.

```python
import eodag

eodag.setup_logging(3)
dag = eodag.EODataAccessGateway()

params = {
    "provider": "cop_ads",
    "collection": "CAMS_SOLAR_RADIATION",
    "time_reference": "universal_time",
    "data_format": "netcdf",
    "sky_type": "clear",
    "time_step": "1day",
    "geometry": [10, 40, 12, 42],
    "altitude": 1,
}
result = dag.search(**params)
dag.download(result[0])
```

PR #2009, opened 2026-01-23
fix: param `ecmwf:date` can be passed also as list

dalpasso:
### Description

`ecmwf:date` is defined as a string parameter but it should be accepted also if passed within a list. In this case the first element of the list is used as date. i.e. both `{"ecmwf:date": "2024-11-18/2024-11-18"}` and `{"ecmwf:date": ["2024-11-18/2024-11-18"]}` are valid search parameters.

### Why is it necessary?

This change is necessary cause `stac-fastapi-eodag` is passing all the search parameters as list of values. This PR is a fix of https://github.com/CS-SI/eodag/pull/1951

### DIFF
--- a/tests/units/test_core.py
+++ b/tests/units/test_core.py
@@ -1905,22 +1905,20 @@ class TestCore(TestCoreBase):
         mock_auth_plugin: mock.Mock,
     ):
         """The date can be passed as string or list"""
-        mock__fetch_data.side_effect = [
+        side_effect = [
             [{"date": ["2025-01-01/2025-01-31"]}],  # constraints
             {},  # form
         ]
-        provider = "wekeo_ecmwf"
-        collection = "AG_ERA5"
-
         params = {
-            "provider": provider,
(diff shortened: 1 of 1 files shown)


tests/units/test_core.py
```diff
@@ -1886,6 +1886,42 @@ def test_list_queryables_dynamic_discover_queryables(
...
+        mock_auth_plugin: mock.Mock,
+    ):
+        """The date can be passed as string or list"""
+        side_effect = [
+            [{"date": ["2025-01-01/2025-01-31"]}],  # constraints
+            {},  # form
+        ]
+        params = {
+            "provider": "wekeo_ecmwf",
+            "collection": "AG_ERA5",
```
sbrunato, line 1914:
```suggestion
            "provider": "cop_ads",
            "collection": "CAMS_GFE_GFAS",
```
AG_ERA5 does not have `ecmwf:date` queryable

Issue #1325, opened 2024-09-24
HTTP-S3 Streaming Download Plugin


Padeanu:
As a user I would want a download plugin that directly forwards HTTP response to a given S3 bucket without using local file system.

Chunk size, target url, source s3 bucket and key, credentials should be configurable.

Example using boto3:

```python
with requests.get(stream_url, stream=True, auth=auth) as response:
    response.raise_for_status()  # Raise an error for bad responses (4xx and 5xx)
    chunk_size = 64 * 1024  # 64kb
    with response.raw as data_stream:
        boto3_client.upload_fileobj(data_stream, bucket, key, Config=boto3.s3.transfer.TransferConfig(multipart_threshold=chunk_size * 2))
```

PR #2021, opened 2026-01-30
feat(core): exposed stream_download method

### LINKED ISSUE #2013: First class support for streaming downloads in EODAG
## What would you like to be added ?

The support for streaming downloads that is already available for use in server mode should also be available for users in the eodag library.

## Why is this needed ?

Streaming downloads would allow users to send the data they retrieved from EODAG to another tool of their choice without having to store on their filesystem first. This feature would enable more powerful and performant data processing pipelines using the EODAG library. 

- the new `stream_download` method will be used in `stac-fastapi-eodag` in place of `_stream_download_dict`
- this should also be demonstrated in a tutorial to upload products/assets on a local S3 created with `minio`: will fix https://github.com/CS-SI/eodag/issues/1325

## How should it be implemented ?

Currently there is a method `plugin.__stream_download_dict()` that is implemented in the download plugins. 
Proposed enhancements:

- [ ] rename `plugin._stream_download_dict` to `plugin.stream_download`
- [ ] rename `plugin._stream_download` to `plugin._raw_stream_download`
- [ ] add `product.stream_download` & `asset.stream_download` methods
- [ ] add a tutorial that uses this method to upload files on a local S3 created with `minio`

No need to have the `stream_download` method exposed through `EODataAccessGateway`.

# Technical note
Add StreamResponse.from_file methode: emulte stream from a file
Warning: core.stream_download return not a real byte stream, but a equivalent of file generator, it's sync, cannot be parallelized. [Enhance StreamResponse behaviour](https://github.com/CS-SI/eodag/issues/2020)

[2026-02-16] Reset work to reapply requirements.

StreamResponse has to be fix to be compliant with ByteIO interface (used by boto3)
MiniIO is not supported anymore, update tutorial, by using rustfs as local s3

_________________________________________________________________________________________________________________________
### Anomaly detected during sphinx compilation:

/home/docs/checkouts/readthedocs.org/user_builds/eodag/checkouts/2021/docs/notebooks/tutos/tuto_download_stream_to_local_s3.ipynb:91: WARNING: Inline substitution_reference start-string without end-string. [docutils]
looking for now-outdated files... none found

> Code extract
```
73  {
74      "cell_type": "code",
75      "execution_count": null,
76      "metadata": {},
77      "outputs": [],
78      "source": [
79          "import boto3\n",
80          "import eodag\n",
81          "\n",
82          "s3_host = \"http://127.0.0.1:9000\"\n",
83          "s3_username = \"rustfs\"\n",
84          "s3_password = \"rustfs\"\n",
85          "s3_region = \"EU\"\n",
86          "s3_bucket = \"test\"\n",
87          "chunk_size = 8388608  # 8Mo\n",
88          "\n",
89          "# Connect to local S3\n",
90          "s3 = boto3.client(\n",
91          "    \"s3\",\n",
92          "    region_name=s3_region,\n",
93          "    aws_access_key_id=s3_username,\n",
94          "    aws_secret_access_key=s3_password,\n",
95          "    endpoint_url=s3_host\n",
96          ")\n",
97          "\n",
98          "# Gather bucklets list\n",
```

### Fix tried:
- rewrites indent space (could bug if there's a tab)
- rename S3_REGION to s3_region (upper case can be interpretated as var by sphinx)
- try 2 spaces indents, 4 spaces indent

### Problem partially identified:
The problem appears to be related to a conflict between Sphinx's templating system (using "|") and Markdown table naming conventions. Depending on the position of certain spaces and underscores ("_"), the system no longer distinguishes between interpretable templates and Markdown syntax.

This would therefore be a problem directly related to the Sphinx interpretation engine, and thus unsolvable. The only solution considered is to integrate the documentation elements block by block, attempting to work around the problem rather than solve it.

pdavid-cssopra:
Fixes #2013
Fixes https://github.com/CS-SI/eodag/issues/1325

### What would you like to be added ?

The support for streaming downloads that is already available for use in server mode should also be available for users in the eodag library.
Why is this needed ?

Streaming downloads would allow users to send the data they retrieved from EODAG to another tool of their choice without having to store on their filesystem first. This feature would enable more powerful and performant data processing pipelines using the EODAG library.
        
the new stream_download method will be used in stac-fastapi-eodag in place of _stream_download_dict

this should also be demonstrated in a tutorial to upload products/assets on a local S3 created with minio: will fix https://github.com/CS-SI/eodag/issues/1325

###  How should it be implemented ?

Currently there is a method plugin.__stream_download_dict() that is implemented in the download plugins. This method should be used in a public method EODataAccessGateway.stream_download(product) similar to EODataAccessGateway.download(product).
Proposed enhancements:
- rename plugin._stream_download_dict to plugin.stream_download
- rename plugin._stream_download to plugin._raw_stream_download
- add product.stream_download & asset.stream_download methods
- add a tutorial that uses this method to upload files on a local S3 created with minio

No need to have the stream_download method exposed through EODataAccessGateway. 

# Technical note

Minio is no longer supported; `rustfs` is suggested as a replacement for tutorials.

Tests update reveals somes bugs:
- [fix(download) get_quicklook result non empty on http 404](https://github.com/CS-SI/eodag/pull/2063)
- [fix(download): product archive wrong filename](https://github.com/CS-SI/eodag/pull/2064)

### DIFF
--- a/tests/units/test_eoproduct.py
+++ b/tests/units/test_eoproduct.py
@@ -24,121 +24,38 @@ import shutil
 import string
 import tempfile
 import time
-import unittest
 import zipfile
-from typing import Optional
 
 import geojson
 import responses
 from lxml import html
-from shapely import geometry, wkt
+from shapely import geometry
 
 from eodag.config import PluginConfig
-from tests import EODagTestCase
+from tests import EODagTestBase
 from tests.context import (
     DEFAULT_SHAPELY_GEOMETRY,
     NOT_AVAILABLE,
     DatasetDriver,
     Download,
     EOProduct,
-    HTTPDownload,
     ProgressCallback,
     mock,
 )
 
-jp = os.path.join
-dirn = os.path.dirname
 
-TEST_RESOURCES_PATH = jp(dirn(__file__), "..", "resources")
-RESOURCES_PATH = jp(dirn(__file__), "..", "..", "eodag", "resources")
-
-
-class TestEOProduct(unittest.TestCase):
+class TestEOProduct(EODagTestBase):
     NOT_ASSOCIATED_COLLECTION = "EODAG_DOES_NOT_SUPPORT_THIS_COLLECTION"
 
     def setUp(self):
         super(TestEOProduct, self).setUp()
         self.output_dir = tempfile.mkdtemp()
-        self.provider = "creodias"
-        self.download_url = (
-            "https://zipper.creodias.eu/download/8ff765a2-e089-465d-a48f-cc27008a0962"
-        )
-        self.local_filename = (
-            "S2A_MSIL1C_20180101T105441_N0206_R051_T31TDH_20180101T124911"
-        )
-        self.local_product_abspath = os.path.abspath(
-            jp(TEST_RESOURCES_PATH, "products", self.local_filename)
-        )
-        self.local_product_as_archive_path = os.path.abspath(
-            jp(
-                TEST_RESOURCES_PATH,
-                "products",
-                "as_archive",
-                "{}.zip".format(self.local_filename),
-            )
-        )
-        self.local_asset_path = os.path.join(
-            TEST_RESOURCES_PATH,
-            "products",
-            "S2A_MSIL1C_20180101T105441_N0206_R051_T31TDH_20180101T124911",
-            "GRANULE",
-            "L1C_T31TDH_A013204_20180101T105435",
-            "IMG_DATA",
-            "T31TDH_20180101T105441_B01.jp2",
-        )
-        self.local_band_file = jp(
-            self.local_product_abspath,
-            "GRANULE",
-            "L1C_T31TDH_A013204_20180101T105435",
-            "IMG_DATA",
-            "T31TDH_20180101T105441_B01.jp2",
-        )
-        # A good valid geometry of a sentinel 2 product around Toulouse
-        self.geometry = wkt.loads(
-            "POLYGON((0.495928592903789 44.22596415476343, 1.870237286761489 "
-            "44.24783068396879, "
-            "1.888683014192297 43.25939191053712, 0.536772323136669 43.23826255332707, "
-            "0.495928592903789 44.22596415476343))"
-        )
-        # The footprint requested
-        self.footprint = {
-            "lonmin": 1.3128662109375002,
-            "latmin": 43.65197548731186,
-            "lonmax": 1.6754150390625007,
-            "latmax": 43.699651229671446,
-        }
-        self.collection = "S2_MSI_L1C"
-        self.platform = "S2A"
-        self.instrument = "MSI"
-        self.provider_id = "9deb7e78-9341-5530-8fe8-f81fd99c9f0f"
-
-        self.eoproduct_props = {
-            "id": "9deb7e78-9341-5530-8fe8-f81fd99c9f0f",
-            "geometry": {
-                "type": "Polygon",
-                "coordinates": [
-                    [
-                        [0.495928592903789, 44.22596415476343],
-                        [1.870237286761489, 44.24783068396879],
-                        [1.888683014192297, 43.25939191053712],
-                        [0.536772323136669, 43.23826255332707],
-                        [0.495928592903789, 44.22596415476343],
-                    ]
-                ],
-            },
-            "collection": self.collection,
-            "constellation": "Sentinel-2",
-            "platform": self.platform,
-            "instruments": [self.instrument],
-            "title": self.local_filename,
-            "eodag:download_link": self.download_url,
-        }
 
     def tearDown(self):
-        super(TestEOProduct, self).tearDown()
         if os.path.isdir(self.output_dir):
             shutil.rmtree(self.output_dir)
         responses.calls.reset()
+        super(TestEOProduct, self).tearDown()
 
     def get_mock_downloader(self):
         """Returns a mock downloader with a default configuration."""
@@ -210,13 +127,13 @@ class TestEOProduct(unittest.TestCase):
         self.assertEqual(geo_interface["type"], "Feature")
         self.assertEqual(
             geo_interface["geometry"],
-            EODagTestCase._tuples_to_lists(geometry.mapping(self.geometry)),
+            EODagTestBase._tuples_to_lists(geometry.mapping(self.geometry)),
         )
         properties = geo_interface["properties"]
         self.assertEqual(properties["eodag:provider"], self.provider)
         self.assertEqual(
             properties["eodag:search_intersection"],
-            EODagTestCase._tuples_to_lists(
+            EODagTestBase._tuples_to_lists(
                 geometry.mapping(product.search_intersection)
             ),
         )
@@ -232,8 +149,8 @@ class TestEOProduct(unittest.TestCase):
                 product.location,
                 product.properties["title"],
                 product.properties["instruments"],
-                EODagTestCase._tuples_to_lists(geometry.mapping(product.geometry)),
-                EODagTestCase._tuples_to_lists(
+                EODagTestBase._tuples_to_lists(geometry.mapping(product.geometry)),
+                EODagTestBase._tuples_to_lists(
                     geometry.mapping(product.search_intersection)
                 ),
                 product.collection,
@@ -245,8 +162,8 @@ class TestEOProduct(unittest.TestCase):
                 same_product.location,
                 same_product.properties["title"],
                 same_product.properties["instruments"],
-                EODagTestCase._tuples_to_lists(geometry.mapping(same_product.geometry)),
-                EODagTestCase._tuples_to_lists(
+                EODagTestBase._tuples_to_lists(geometry.mapping(same_product.geometry)),
+                EODagTestBase._tuples_to_lists(
                     geometry.mapping(same_product.search_intersection)
                 ),
                 same_product.collection,
@@ -537,7 +454,8 @@ class TestEOProduct(unittest.TestCase):
                     self.eoproduct_props,
                     **{"eodag:download_link": "http://example.com/foobar.zip"},
                 )
-            )
+            ),
+            extract=False,
         )
 
         # Download, and check that the mocked request was properly called.
@@ -573,7 +491,8 @@ class TestEOProduct(unittest.TestCase):
                     "title": "asset title",
                     "type": "image/jp2",
                 }
-            }
+            },
+            extract=False,
         )
 
         # Download, and check that the mocked request was properly called.
@@ -645,7 +564,8 @@ class TestEOProduct(unittest.TestCase):
                         "title": "".join(random.choices(string.ascii_letters, k=10)),
                     },
                 )
-            )
+            ),
+            extract=False,
         )
         progress_callback = ProgressCallback()
 
@@ -733,7 +653,8 @@ class TestEOProduct(unittest.TestCase):
                             "otherProperty": "%(/%s/neither/resolved",
                         },
                     )
-                )
+                ),
+                extract=False,
             )
             self.assertEqual(downloadable_product.location, "%(257B/cannot/be/resolved")
             self.assertEqual(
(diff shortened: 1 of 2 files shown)


eodag/api/core.py
sbrunato, line 1: feature request description updated: please apply changes

tests/units/test_eoproduct.py
```diff
@@ -410,6 +410,80 @@ def test_eoproduct_download_http_extract(self):
...
+            {
+                "foo": {
+                    "href": "https://example.com/asset/foo.tif",
+                    "title": "foo asset",
+                    "type": "image/tiff",
+                    "proj:shape": [3, 343, 343],
+                    "mgrs:utm_zone": "also-bad-formatted",
+                }
+            }
+        )
```
sbrunato: what is it for ?

tests/units/test_eoproduct.py
```diff
@@ -410,6 +410,80 @@ def test_eoproduct_download_http_extract(self):
...
+        product.properties["grid:code"] = "MGRS-31TCJ"
+        product.properties["eo:cloud_cover"] = "bad-formatted"
+        product.assets.update(
+            {
+                "foo": {
+                    "href": "https://example.com/asset/foo.tif",
+                    "title": "foo asset",
+                    "type": "image/tiff",
+                    "proj:shape": [3, 343, 343],
+                    "mgrs:utm_zone": "also-bad-formatted",
```
sbrunato: why these properties ?

tests/units/test_eoproduct.py
```diff
@@ -410,6 +410,80 @@ def test_eoproduct_download_http_extract(self):
...
+                }
+            }
+        )
+
+        # Download
+        asset_stream = product.assets["foo"].stream_download()
+        self.assertEqual(asset_stream.headers["content-type"], "image/tiff")
+        self.assertEqual(asset_stream.headers["content-length"], "1")
```
sbrunato: a content could be downloaded here

eodag/utils/streamresponse.py
```diff
@@ -0,0 +1,203 @@
...
+    _size: Optional[int] = field(default=None, repr=False, init=False)
+    headers: dict[str, str] = field(default_factory=dict)
+    media_type: Optional[str] = None
+    status_code: Optional[int] = None
+    arcname: Optional[str] = None
+
+    @staticmethod
+    def from_file(
+        file_path: str, on_complete: Optional[Callable] = None
+    ) -> "StreamResponse":
```
sbrunato: please test this method

eodag/utils/streamresponse.py
```diff
@@ -0,0 +1,203 @@
...
+        """return a file as a byte stream"""
```
sbrunato: please complete docstring

eodag/utils/streamresponse.py
```diff
@@ -0,0 +1,203 @@
...
+import types
+from dataclasses import dataclass, field
+from glob import glob
+from typing import Callable, Iterable, Iterator, Mapping, Optional
+
+
+class StreamResponseContent(Iterable[bytes]):
+    """
+    ByteIO minimal compatibility, used by boto3.upload_fileobj that's
+    usually expect BytesIO object, not Iterable[bytes]
```
sbrunato, line 10:
```suggestion
    ``io.BytesIO`` minimal compatibility for ``StreamResponse.content``.
    
    Used by methods like ``boto3.upload_fileobj`` expecting ``io.BytesIO`` object implementing a ``read`` method.
```

eodag/utils/streamresponse.py
```diff
@@ -0,0 +1,203 @@
...
+            signal.signal(signal.SIGTERM, signal_handler)
+
+    def __init__(self, content: Iterable[bytes]):
+        self.iterator = iter(content)
+        self.buffer: bytes = b""
+
+    def __iter__(self) -> Iterator[bytes]:
+        return self.iterator
+
+    def read(self, size=-1) -> bytes:
```
sbrunato, line 46: please test using tuto-like s3 upload with `moto`

eodag/api/product/_product.py
```diff
@@ -481,6 +481,54 @@ def download(
...
+        wait: float = DEFAULT_DOWNLOAD_WAIT,
+        timeout: float = DEFAULT_DOWNLOAD_TIMEOUT,
+        **kwargs: Unpack[DownloadConf],
+    ) -> StreamResponse:
+        """Download as StreamResponse the EO product using the provided download plugin and the
+        authenticator if necessary.
+
+        The actual download of the product occurs only at the first call of this
+        method. A side effect of this method is that it changes the ``location``
+        attribute of an EOProduct, from its remote address to the local address.
```
sbrunato:
```suggestion
```

eodag/api/product/_product.py
```diff
@@ -481,6 +481,54 @@ def download(
...
+        :param byte_range: (optional) Tuple of first index / last index byte to read
+        :param compress: (optional) "zip", "raw", "auto"
+        :param wait: (optional) If download fails, wait time in minutes between
+                     two download tries
+        :param timeout: (optional) If download fails, maximum time in minutes
+                        before stop retrying to download
+        :param kwargs: `output_dir` (str), `extract` (bool), `delete_archive` (bool)
+                        and `dl_url_params` (dict) can be provided as additional kwargs
+                        and will override any other values defined in a configuration
+                        file or with environment variables.
```
sbrunato:
```suggestion
        :param kwargs: additional kwargs like `dl_url_params` (dict) can be provided
                        and will override any other values defined in a configuration
                        file or with environment variables.
```

docs/notebooks/tutos/tuto_stream_download_to_s3.ipynb
```diff
@@ -0,0 +1,576 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "<div class=\"alert alert-info\">\n",
+    "\n",
+    "Note\n",
+    "\n",
+    "Available assets can also be downloaded as stream using [EOProduct.stream_download()](../../api_reference/assets.rst#eodag.api.product._assets.Asset.stream_download):\n",
```
sbrunato:
```suggestion
    "Available assets can also be downloaded as stream using [Asset.stream_download()](../../api_reference/assets.rst#eodag.api.product._assets.Asset.stream_download):\n",
```

tests/units/test_eoproduct.py
pdavid-cssopra: Leftover from a poorly cleaned second test, I'm removing it.

tests/units/test_eoproduct.py
pdavid-cssopra: That's the case
```
# Download to tmp directory
        filepath = os.path.join(self.output_dir, asset_stream.filename)
        with open(filepath, "wb") as fp:
            for chunk in asset_stream.content:
                fp.write(chunk)
```
Global emultation is unconsistent, even if has headers like content length = 1, mock not serve real file

eodag/utils/streamresponse.py
pdavid-cssopra: Out-of-scope remainder, removed.

eodag/utils/streamresponse.py
pdavid-cssopra: Out-of-scope remainder, removed.

eodag/utils/streamresponse.py
pdavid-cssopra: No code section is specifically designed for S3 uploads, so there is no code to which an S3 upload test can be attached. The upload in the tutorial is only a way to test the upload, but is not part of the functional scope covered here.

As a reminder, streaming response takes a generator or a file as its source, but never a downstream stream.

sbrunato: If the suggestion does not suit you, do it another way, but please test `StreamResponse.read` method that is 0% covered

pdavid-cssopra: Understood

tests/units/test_eoproduct.py
sbrunato: Several issues here:
- `content-length` should not be 1. Problem fixed by https://github.com/CS-SI/eodag/pull/2054
- downloaded file name is not correct (comes from `_dummy_downloadable_product`). Please adapt mock for asset in this test

pdavid-cssopra: The method for emulating external requests is inadequate for the required tests, which demand sequencing multiple calls. Lacking the time to properly recreate the emulations, the problem will be solved with an unsightly workaround.

Regarding HTTP headers, the IEEE retains the HTTP 1.0 header, which is the most restrictive of the three current standard versions (1.0, 1.1, 2.0), in a CamelCase-like format, such as "Content-Length".

It's reassuring that the code has a safeguard, but not that it compensates for inconsistencies in code writing. All fields have been standardized to a common naming convention.

pdavid-cssopra: Here is a representative example of what HTTP emulation mechanics for testing could look like. 
[http_mock.py](https://github.com/user-attachments/files/25546718/http_mock.py)

sbrunato: like already discussed, we are using `responses` in tests to handle multiple calls. Please use it instead of rewriting an alternative

tests/mocks/responsefile.py
```diff
@@ -0,0 +1,123 @@
+import datetime
+import mimetypes
+import os
+import signal
+from typing import Union
+
+
+class ResponseFile:
```
sbrunato: please use `responses` library. A file content can be passed as body if needed https://github.com/getsentry/responses?tab=readme-ov-file#response-parameters

pdavid-cssopra: this class replace an existing code in test:

tests/units/test_download_plugins.py:230
```
    def _download_response_archive(self):
        class Response:
            """Emulation of a response to requests.get method for a zipped product"""

            def __init__(response):
                # Using a zipped product file
                with open(self.local_product_as_archive_path, "rb") as fh:
                    response.__zip_buffer = io.BytesIO(fh.read())
                cl = response.__zip_buffer.getbuffer().nbytes
                response.headers = CaseInsensitiveDict(
                    {
                        "content-length": cl,
                        "content-disposition": "attachment; filename=foobar.zip",
                    }
                )
                response.url = "http://foo.bar"

            def __enter__(response):
                return response

            def __exit__(response, *args):
                pass

            def iter_content(response, **kwargs):
                with response.__zip_buffer as fh:
                    while True:
                        chunk = fh.read(kwargs["chunk_size"])
                        if not chunk:
                            break
                        yield chunk

            def raise_for_status(response):
                pass

            def close(response):
                pass

        return Response()
```

The class has been generalized to allow for extending tests without rewriting existing ones. The file has been isolated into a separate file for cleanliness and dependency isolation.

The response library does not allow overloading the response object of a request, and is therefore not suitable here.

tests/units/test_eoproduct.py
pdavid-cssopra: responses librairie does not expose overloadable classes and not allow a way to adapt protocol behaviour. It's unfortunatly not enough to manage custom case like streaming, in particular when eodag does not go through the standard stream transfer, and does not expose a generic generator object.

sbrunato: we do not need to overload responses classes, neither to adapt protocol behavior. Please test following this example https://github.com/CS-SI/eodag/pull/2062

tests/mocks/responsefile.py
sbrunato: this manual mechanism was also not needed, I just removed it in https://github.com/CS-SI/eodag/pull/2062.

Please adapt your tests following this example

pdavid-cssopra: conflict appears between response librairie and native mocking from EodagTestCase.
Heritance from EodagTestCase to TestEoProduct had to be removed to implement requested tests update.

tests/units/test_eoproduct.py
pdavid-cssopra: conflict appears between response librairie and native mocking from EodagTestCase.
Heritance from EodagTestCase to TestEoProduct had to be removed to implement requested tests update.

Issue #1922, opened 2025-11-06
collections summaries


sbrunato:
Move collections properties (`instruments`, `platform`, `constellation`, ...) to `summaries` to align with STAC formatting

PR #2052, opened 2026-02-16
feat(core): STAC summaries in Collections

anesson-cs:
Linked to this [PR](https://github.com/CS-SI/eodag/pull/1940) and fix this [issue](https://github.com/CS-SI/eodag/issues/1922).

eodag collection model now inherits `Collection` model from library `stac_pydantic` to be fully STAC-compliant. Some fields are adapted to eodag to deal with cases where values are not provided in the input, especially when fields are mandatory. Field `keywords` is adapted for typing reason.

`model_dump()` and `model_dump_json()` methods are overwritten to get unset values and display 'stac_extensions' property if wanted. As it takes too long to set `stac_extensions` for a big amount of collections, a parameter `display_extensions` has been created to display it with these methods when needed.

The wrap validator has been updated to handle errors of an element in a dictionary or list field as these fields are now accepted in the model.

Some validators have been created to deal with field `summaries`, which is a dictionary whose key-value pairs are checked in the model. Pairs set directly in model fields have priority over the ones set in field `summaries`.

`proprietary` is replaced by `other` for collection metadata `license` which still had this value in some collections.

The default value of collection metadata `instruments` is now set to `None` as it is not mandatory.


sbrunato (approved):

eodag/api/core.py
sbrunato, line 1: rebase needed

sbrunato (dismissed):

Issue #2060, opened 2026-02-26
Search validation error when using ecmwf:longitude and ecmwf:latitude


pdavid-cssopra:
### What happened ?

The ECMWF parameter location has the following structure:
```
{
    "location" : {
        "longitude": "10",
        "latitude": "40"
    }
}
```
Using it with the prefix ecmwf: causes the validation error:
```
eodag.utils.exceptions.ValidationError: The location must contains the latitude and the longitude
```
It's possible to search and download the product if you omit the prefix `ecmwf: prefix

### What did you expect to happen ?

It should be possible to use the parameters `location`, `longitude`, `latitude` with and without the ecmwf: prefix.

###  How can we reproduce it (as minimally and precisely as possible)?

```
from pprint import pprint
from eodag.api.core import EODataAccessGateway
from eodag.utils.logging import setup_logging

setup_logging(3)
dag = EODataAccessGateway()

params = {
    "provider": "cop_ads",
    "collection": "CAMS_SOLAR_RADIATION",
    "ecmwf:sky_type": "clear",
    "ecmwf:data_format": "netcdf",
    "ecmwf:time_reference": "universal_time",
    "ecmwf:time_step": "1day",
    "ecmwf:date": "2026-02-10/2026-02-10",
    "ecmwf:location": {"ecmwf:longitude": 11, "ecmwf:latitude": 41},
    "ecmwf:altitude": "1",
}
res = dag.search(**params, raise_errors=True)
dag.download(res[0])
```

### Anything else we need to know ? Any hints on how to fix the problem?

The exception is raised by eodag.utils.__init__#get_geometry_from_ecmwf_location

PR #2061, opened 2026-02-26
fix(search): search validation error when using ecmwf:longitude and e…

pdavid-cssopra:
From [#2060](https://github.com/CS-SI/eodag/issues/2060)
### What happened ?

The ECMWF parameter location has the following structure:
```
{
    "location" : {
        "longitude": "10",
        "latitude": "40"
    }
}
```
Using it with the prefix ecmwf: causes the validation error:
```
eodag.utils.exceptions.ValidationError: The location must contains the latitude and the longitude
```
It's possible to search and download the product if you omit the prefix `ecmwf: prefix

### What did you expect to happen ?

It should be possible to use the parameters `location`, `longitude`, `latitude` with and without the ecmwf: prefix.

###  How can we reproduce it (as minimally and precisely as possible)?

```
from pprint import pprint
from eodag.api.core import EODataAccessGateway
from eodag.utils.logging import setup_logging

setup_logging(3)
dag = EODataAccessGateway()

params = {
    "provider": "cop_ads",
    "collection": "CAMS_SOLAR_RADIATION",
    "ecmwf:sky_type": "clear",
    "ecmwf:data_format": "netcdf",
    "ecmwf:time_reference": "universal_time",
    "ecmwf:time_step": "1day",
    "ecmwf:date": "2026-02-10/2026-02-10",
    "ecmwf:location": {"ecmwf:longitude": 11, "ecmwf:latitude": 41},
    "ecmwf:altitude": "1",
}
res = dag.search(**params, raise_errors=True)
dag.download(res[0])
```

### Anything else we need to know ? Any hints on how to fix the problem?

The exception is raised by eodag.utils.__init__#get_geometry_from_ecmwf_location

### DIFF
--- a/eodag/plugins/search/build_search_result.py
+++ b/eodag/plugins/search/build_search_result.py
@@ -21,9 +21,10 @@ import hashlib
 import logging
 import re
 from collections import OrderedDict
+from collections.abc import Callable
 from datetime import datetime, timedelta
 from types import MethodType
-from typing import TYPE_CHECKING, Annotated, Any, Optional
+from typing import TYPE_CHECKING, Annotated, Any, Optional, Tuple
 from urllib.parse import quote_plus, unquote_plus
 
 import geojson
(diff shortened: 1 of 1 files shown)


sbrunato:
we do not want to have `ecmwf:` prefix inside parameters/properties objects (stay with `"ecmwf:location": {"longitude": 11, "latitude": 41}`)

PR #2103, opened 2026-03-21
build: updated package discovery pattern

avalentino:
Fix setuptools package discovery.
Currently the wheel package contains unexpected contents:
```
eodag-4.0.1-py3-none-any
- docs
- eodag
- eodag-4.0.1.dist-info
- utils
```
The patch changes the package discovery configuration to include only "eodag*"

Thank you!


avalentino:
The failure seems to be due to a missing token.
It is not related to this PR.

sbrunato (approved):

PR #2112, opened 2026-03-25
feat(plugins): basic auth support for OIDC-protected assets and 404 error handling

cauriol:
This PR extends OIDC authentication handling to support Basic authorization for protected asset access and improves handling for 404 errors in `HTTPDownload`.

Changes:
- extend CodeAuthorizedAuth usage to support Basic authorization headers for protected asset access
- protect OIDC authentication token retrieval with a lock to avoid concurrent token refresh issues
- return a NotAvailableError if only one asset is downloaded and the provider returns a 404 response


alambare:
@cauriol, instead of introducing a new method `get_auth_headers` in plugin configs, we should leverage the `product._get_storage_options` from EODAG cube: https://github.com/CS-SI/eodag-cube/blob/develop/eodag_cube/api/product/_product.py#L121.
This method should probably be refactored to not activate order and become a public interface.

@sbrunato can we get your opinion here ?

sbrunato:
> @cauriol, instead of introducing a new method `get_auth_headers` in plugin configs, we should leverage the `product._get_storage_options` from EODAG cube: https://github.com/CS-SI/eodag-cube/blob/develop/eodag_cube/api/product/_product.py#L121. This method should probably be refactored to not activate order and become a public interface.
> 
> @sbrunato can we get your opinion here ?

@alambare I agree. `EOProduct._get_storage_options` can be moved from `eodag-cube` to `eodag` and publicly exposed

eodag/api/product/_product.py
```diff
@@ -525,6 +535,88 @@ def stream_download(
...
+
+        return {"path": url}
+
+    def request_asset(
+        self,
+        url: str,
+    ) -> requests.Response:
+        """Perform a GET request to the given URL using product's authentication headers."""
+        headers = self.get_storage_options().get("headers", {})
+        return requests.get(url, headers=headers, stream=True)
```
alambare, line 691: we should use the stream download method from EODAG download plugins instead of creating a new method.

eodag/api/product/_product.py
```diff
@@ -525,6 +535,88 @@ def stream_download(
...
+        import fsspec  # type: ignore[import-untyped]
+
+        headers = self.get_storage_options().get("headers", {})
+        mapper = fsspec.get_mapper(
+            base_url,
+            client_kwargs={"headers": headers, "trust_env": False},
+        )
+
+        if ".zmetadata" in mapper:
+            meta = json.loads(mapper[".zmetadata"])
```
alambare: Use `orjson` instead of `json`.

eodag/api/product/_product.py
```diff
@@ -525,6 +535,88 @@ def stream_download(
...
+            return [".zmetadata", *meta["metadata"].keys()]
+
+        # TODO: Support Zarr v3 when test data becomes available.
+        # Zarr v2 uses `.zmetadata`, while Zarr v3 exposes `zarr.json`.
+        # The implementation should be straightforward once we can validate it
+        # against real examples.
+        raise ValueError(f"No Zarr metadata file found at {base_url}")
```
alambare: With Destination Earth, we do have a zarr v3 store: DanubeHis data. I don't understand what is blocking here.

eodag/api/product/_product.py
```diff
@@ -525,6 +535,88 @@ def stream_download(
...
+
+    def request_asset(
+        self,
+        url: str,
+    ) -> requests.Response:
+        """Perform a GET request to the given URL using product's authentication headers."""
+        headers = self.get_storage_options().get("headers", {})
+        return requests.get(url, headers=headers, stream=True)
+
+    def list_zarr_files_from_metadata(
```
alambare: In zarr, we have `keys` and not `files`. I suppose you can rename the method to something like `list_zarr_keys` ?

eodag/api/product/_product.py
```diff
@@ -525,6 +535,88 @@ def stream_download(
...
+        """
+        auth = self.downloader_auth.authenticate() if self.downloader_auth else None
+        if self.downloader is None:
+            return {}
+
+        # default url and headers
+        try:
+            url = self.assets[asset_key]["href"] if asset_key else self.location
+        except KeyError as e:
+            raise DatasetCreationError(f"{asset_key} not found in {self} assets") from e
```
alambare: You are not creating a xarray here. The exception class is not accurate.

pyproject.toml
```diff
@@ -59,6 +59,9 @@ dependencies = [
     "typing_extensions >= 4.8.0",
     "urllib3",
     "zipstream-ng",
+    "fsspec",
+    "aiohttp",
+    "requests"
```
alambare: `requests` is already a dependency in EODAG.

eodag/api/product/_product.py
```diff
@@ -525,6 +535,88 @@ def stream_download(
...
+        base_url: str,
+    ) -> list[str]:
+        """List file paths from a Zarr store metadata file."""
+        import fsspec  # type: ignore[import-untyped]
+
+        headers = self.get_storage_options().get("headers", {})
+        mapper = fsspec.get_mapper(
+            base_url,
+            client_kwargs={"headers": headers, "trust_env": False},
+        )
```
alambare: I am not sure about introducing fsspec in EODAG library.
The library is a wrapper above multiple backends. To use it, you need to bring backends to support all the filesystems: `s3fs` for S3, `gcsfs` for google cloud storage and `adlfs` for azure blob storage.
Most of the times, users will store zarr files on a S3 compatible store but some may store them on google or azure stores. 
Maybe we don't want to support google / azure stores ? or in an optional dependencies branch ?

My opinion:
1. we do not support google and azure yet. Maybe once if we replace `boto3` by `obstore`.
2. we do not bring `fsspec` in EODAG but instead call `boto3` to get the list of zarr keys. If a dedicated function is needed to get the list of keys with boto3, it should live in `eodag/utils/s3.py`.

@sbrunato what do you think?

alambare (changes requested):

eodag/api/product/_product.py
sbrunato: i think we should not import `fsspec` directly in `eodag` and also, not sure if we should handle zarr subsetting from eodag.

Having storage options is great and should be useful to open zarr from xarray or eodag-cube

eodag/api/product/_product.py
cauriol: I put AddressNotFound now

eodag/api/product/_product.py
sbrunato, line 1: `get_storage_options` implementation should be done in a separate PR

eodag/api/product/_product.py
sbrunato: method to remove

eodag/resources/providers.yml
```diff
@@ -4702,6 +4702,51 @@
     login_form_xpath: //form[@id='kc-form-login']
     authentication_uri_source: login-form
 ---
+!provider # MARK: desp_cache
+  name: desp_cache
```
sbrunato: Provider name and description need to be updated. "Cache" does not seem appropriate in name and description.

eodag/resources/providers.yml
sbrunato, line 1: Provider and its registration are not documented

sbrunato (changes requested):

eodag/api/product/_product.py
jlahovnik: has been extracted to https://github.com/CS-SI/eodag/pull/2276

eodag/resources/providers.yml
jlahovnik: After further consideration I think we cannot add this provider in EODAG because it the data is in internal FDP which is not publicly exposed. But we can add the collections to the dedl provider and we should keep the changes for the auth plugin.

eodag/api/product/_product.py
jlahovnik: method removed and code in stac-fastapi-eodag re-written to use stream_download

jlahovnik:
[desp_cache.txt](https://github.com/user-attachments/files/30544992/desp_cache.txt)
[internal_fdp.txt](https://github.com/user-attachments/files/30545022/internal_fdp.txt)
configs that can be used for testing

sbrunato (approved):

PR #2121, opened 2026-03-30
feat(core): new (de)serialization methods

sbrunato:
Adds new (de)serialization methods:

- `EOProduct.from_dict()`
- `EOProduct.from_file()`
- `SearchResult.from_dict()`
- `SearchResult.from_file()`
- `SearchResult.as_dict()`

If `dag` parameter is passed to deserialization methods, they will try to register downloader using several import methods (from serialized, from eodag-server, from known provider, from unknown provider), like in `EODataAccessGateway.import_stac_items`.

Serialize methods now accept a `skip_invalid` parameter (`True` by default), to choose whether to skip properties whose values are not valid according to the STAC specification.

Deprecated methods:

- `EOProduct.from_geojson()` (use `from_dict` instead)
- `SearchResult.from_geojson()` (use `from_dict` instead)
- `SearchResult.as_geojson_object()` (use `as_dict` instead)

`unregistered_product_from_item` moved to `eodag.utils.deserialize`.


alambare (approved):
Looks good to me

PR #2139, opened 2026-04-09
refactor: quotas errors handling and order-status message

jlahovnik:
- handle 429 error response returned by provider
- use order status as status message if no message in provider response


eodag/utils/exceptions.py
sbrunato, line 1: @jlahovnik , is the refactor introduced in my commit ok for you ?

jlahovnik: Yes, the changes are OK for me.

sbrunato (approved):

PR #2148, opened 2026-04-14
fix(queryables): required parameters, defaults and allowed combinations

dalpasso:
This PR is fixing how the queryables are calculated from provider [form file](https://eodag.readthedocs.io/en/latest/plugins.html#eodag.config.PluginConfig.DiscoverQueryables.form_url).

# Queryables available values

If no constraint is matching a given parameter, then look at the form file:

- if the keyword is required by the form but does not match any combination with associated parameters, then do not show it as queryable
- if the keyword is _not_ required by form, then any value is accepted for it.
 
In the latter case, there may be no constraint at all containing that keyword because any value is anyway accepted.

# Queryables by form

- the default value is set with the following priority:
  - input default values > form details default value > no default value
- a parameter is not allowed and excluded from the queryables if the following conditions are true:
  - the keyword is required by form;
  - the keyword is used by some constraint;
  - the list of available values is empty.

# Test

```python
from eodag.utils.exceptions import ValidationError
from typing import get_args

dag = EODataAccessGateway()

# --------------------------------------------------------------------------------------------------
# 1. Parameter is required but does not match allowed combinations with given associated parameters

# parameter 'level' should be required for given collection
queryables = dag.list_queryables(
    provider="cop_cds", collection="CMIP6_CLIMATE_PROJECTIONS"
)
assert get_args(queryables["ecmwf_level"])[1].is_required(), "'level' should be required"

# check that 'level' is not queryable with these parameters
params = {
    "provider": "cop_cds",
    "collection": "CMIP6_CLIMATE_PROJECTIONS",
    "experiment": "ssp2_4_5",
    "model": "access_cm2",
    "month": ["01"],
    "temporal_resolution": "monthly",
    "variable": "near_surface_air_temperature",
    "year": ["2023"],
}
queryables = dag.list_queryables(**params)
assert "ecmwf_level" not in queryables, "'level' should not be queryable"

# validatation should fail if we search with 'level' parameter
try:
    dag.search(**params, level=["1"], raise_errors=True)
except ValidationError as e:
    # expected because 'level' is not queryable
    assert "cannot be used with this combination of parameters" in str(
        e
    ), "Unexpected error message: {}".format(e)
else:
    raise AssertionError("Search should have failed because 'level' is not queryable")

# must succeed without 'level' parameter
dag.search(**params, raise_errors=True)

# --------------------------------------------------------------------------------------------------
# 2. Parameters are not required, so any value should be accepted

# must succeed: `pressure_level` and `model_level` are not required by form: any value is accepted
# note: the constraint file doesn't contain any constraint matching *all* these parameters,
# but this is fine cause any value is accepted
params = {
    "provider": "cop_ads",
    "collection": "CAMS_GAC_FORECAST",
    "variable": ["ammonium_aerosol_mass_mixing_ratio"],
    "pressure_level": ["1"],
    "model_level": ["1"],
    "date": "2025-12-11/2025-12-11",
    "time": ["00:00"],
    "leadtime_hour": ["0"],
    "type": ["forecast"],
    "data_format": "grib",
    "area": [42, 10, 40, 12],
}
dag.search(**params, raise_errors=True)

```

### DIFF
--- a/eodag/config.py
+++ b/eodag/config.py
@@ -211,10 +211,12 @@ class PluginConfig(yaml.YAMLObject):
         result_type: str
         #: JsonPath to retrieve the queryables from the provider result
         results_entry: str
-        #: :class:`~eodag.plugins.search.base.Search` URL of the constraint file used to build queryables
-        constraints_url: str
         #: :class:`~eodag.plugins.search.base.Search` Key in the json result where the constraints can be found
         constraints_entry: str
+        #: :class:`~eodag.plugins.search.base.Search` URL of the constraint file used to build queryables
+        constraints_url: str
+        #: :class:`~eodag.plugins.search.base.Search` URL of the form file used to build queryables
+        form_url: str
 
     class CollectionSelector(TypedDict, total=False):
         """Define the criteria to select a collection in :class:`~eodag.config.PluginConfig.DynamicDiscoverQueryables`.
--- a/eodag/plugins/search/build_search_result.py
+++ b/eodag/plugins/search/build_search_result.py
@@ -795,7 +795,7 @@ class ECMWFSearch(PostJsonSearch):
             available_values = self.available_values_from_constraints(
                 constraints,
                 non_empty_formated,
-                form_keywords=[f["name"] for f in form],
+                form,
             )
 
             # Pre-compute the required keywords (present in all constraint dicts)
@@ -878,7 +878,7 @@ class ECMWFSearch(PostJsonSearch):
         self,
         constraints: list[dict[str, Any]],
         input_keywords: dict[str, Any],
-        form_keywords: list[str],
+        form: list[dict[str, Any]],
     ) -> dict[str, list[str]]:
         """
         Filter constraints using input_keywords. Return list of available queryables.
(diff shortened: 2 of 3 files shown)


jlahovnik (approved):

sbrunato:
> must fail because not constraint is matching the given params and all keywords are required by form

For the first case, things are confusing. The error raised is:
```
eodag.utils.exceptions.ValidationError: ("ecmwf:level=['1'] is not available with ecmwf:temporal_resolution=monthly, ecmwf:experiment=ssp2_4_5, ecmwf:variable=near_surface_air_temperature. No value allowed.", {'variable', 'temporal_resolution', 'experiment', 'ecmwf:level'})
```
But `ecmwf:level` appears in queryables with these parameters:
```py
>>> dag.list_queryables(**params)["ecmwf_level"]
typing.Annotated[list, FieldInfo(annotation=NoneType, required=False, default=None, alias_priority=2, validation_alias=AliasChoices(choices=['ecmwf:level', 'level']), serialization_alias='ecmwf:level', title='Level')]
```

sbrunato (changes requested):
please remove the not-allowed parameter from queryables, and add a test for this new mechanism

sbrunato (approved):

PR #2151, opened 2026-04-16
fix(plugins): remove EOIAMAuth lazy login causing silent errors

### LINKED ISSUE #2145: eoiam silent auth error
Using `eocat` provider, if we perform:

1. search and download on a collection on which user already set his consent, it succeeds.
2. then on the same session, search and download on a collection on which user did not set his consent, it silently fails.

We should check if download request in `2.` returns a html content, and handle it

timothey-cs:
Fixes #2145

### DIFF
--- a/eodag/plugins/authentication/eoiam.py
+++ b/eodag/plugins/authentication/eoiam.py
@@ -185,12 +185,14 @@ class _EOIAMSessionAuth(AuthBase):
         We use the session's get/post to ensure login happens if needed.
         """
         session = self.auth_plugin.session
+        try:
+            resp = session.get(request.url, allow_redirects=True)
+            if "Earth Observation Identity and Access Management System" in resp.text:
+                resp = self.auth_plugin._login_from_html(resp.text, request.url)
 
-        resp = session.get(request.url, allow_redirects=True)
-        if "Earth Observation Identity and Access Management System" in resp.text:
-            self.auth_plugin._login_from_html(resp.text, req_url=request.url)
+            # Copy cookies from session to the request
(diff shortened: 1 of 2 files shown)


eodag/plugins/authentication/eoiam.py
```diff
@@ -187,12 +186,9 @@ def __call__(self, request):
...
         session = self.auth_plugin.session
 
-        # Lazy login
-        if not self.auth_plugin._logged_in:
-            resp = session.get(request.url, allow_redirects=True)
-            if "Earth Observation Identity and Access Management System" in resp.text:
-                self.auth_plugin._login_from_html(resp.text, req_url=request.url)
-            self.auth_plugin._logged_in = True
+        resp = session.get(request.url, allow_redirects=True)
+        if "Earth Observation Identity and Access Management System" in resp.text:
```
sbrunato: silent error described in issue still happens: check html anchor in content instead of "Earth Observation Identity and Access Management System"

sbrunato (changes requested):

tests/units/test_auth_plugins.py
```diff
@@ -942,29 +944,6 @@ def test_plugins_auth_eoiam_authenticate_returns_authbase(self):
         auth = auth_plugin.authenticate()
         self.assertIsInstance(auth, AuthBase)
 
-    @responses.activate
-    def test_plugins_auth_eoiam_authenticate_no_login_required(self):
```
sbrunato, line 946: can't you keep that test by mocking `_login_from_html` and check that it is not called ?

sbrunato (changes requested):

sbrunato (approved):

PR #2158, opened 2026-04-24
feat: store `federation` metadata in SQLite

dalpasso:
**Important: this PR uses the branch `db-for-collections-storage` as base. Leaving as draft until this branch is merged.**

Add the extension defined in https://github.com/Open-EO/openeo-api/blob/master/extensions/federation/README.md and store it in SQLite database.

This is a preparatory change to extend the fields within the `federation` metadata in the STAC server (https://github.com/CS-SI/stac-fastapi-eodag/pull/108).


alambare:
I don't understand why do we need to add a `federation` column in the database.
For what I understood, it is purely an informational field. We should probably keep it in the `content` column.

dalpasso:
> I don't understand why do we need to add a `federation` column in the database. For what I understood, it is purely an informational field. We should probably keep it in the `content` column.

It's in a separate column to avoid federation metadata to be overwritten by subsequent `upsert_collections()` calls in the STAC Server.

PR #2150, opened 2026-04-16
refactor: dates parsing and handling harmonization

sbrunato:
Harmonizes dates parsing and handling, and uses common methods from [eodag.utils.dates](https://eodag.readthedocs.io/en/latest/api_reference/utils.html#module-eodag.utils.dates)

PR #2159, opened 2026-04-27
fix: datetime interval parsing and normalize UTC output format

cauriol:
An issue occurred when using datetime intervals such as:
http://localhost:8000/collections/S1_SAR_GRD/items?datetime=2023-01-01T00:00:00Z/2023-02-01T00:00:00Z&filter=federation:backends=%27planetary_computer%27

This PR fixes the parsing of `/`-separated datetime ranges and normalizes all outputs to ISO 8601 format using the `Z` suffix for UTC.


jlahovnik (approved):

sbrunato:
already included in https://github.com/CS-SI/eodag/pull/2150

PR #2165, opened 2026-04-29
fix: do not pop feature

cauriol:
When doing a search on dedt_lumi or dedt_mn5 with feature, some parameters were mising. 
for exemple if we search with:
```
feature={
        "type": "timeseries",
        "points": [
                    [
                        -9.10,
                        38.78
                    ]
                ],
                "time_axis": "step",
                "range": {
                    "start": 0,
                    "end": 360
                },
                "axes": [
                    "latitude",
                    "longitude"
                ]
	}
```

We got this in the order: 
```
"feature": {
        "points": [
            [
                -9.1,
                38.78
            ]
        ],
        "type": "position"
    }
```
the parameter time_axis, range and axes were missing. 
By removing the params.pop("feature") we keep all the parameters.


sbrunato:
fixed by #2167

PR #2171, opened 2026-05-01
fix(mapping): normalize bands when product has raster:bands but no eo:bands

lhoupert:
## Summary

Fixes #2169.

`normalize_bands` (introduced in #2050) raises an `IndexError` when a STAC asset has
`raster:bands` but no `eo:bands` (e.g. COP-DEM GLO-30 tiles from `earth_search`).
The exception is silently swallowed by `_do_search`, causing searches to return 0 products
with no warning.

## Root cause

When only `raster:bands` is present, `processed_bands` is empty at the start of the
migration loop. The unconditional `band = processed_bands[index]` on line 1870 crashes
immediately:

```python
for item in bands["raster:bands"]:
    band = processed_bands[index]           # ← IndexError: list is empty
    ...
    if index < len(processed_bands):
        processed_bands[index] = band
    else:
        processed_bands.append(band)        # ← intended fallback, unreachable
```

## Fix

Mirror the guard already present on the write side onto the read side (one line change):

```python
band = processed_bands[index] if index < len(processed_bands) else {}
```

## Quick validation

Patch the installed file and run the search — should go from 0 to 29 products:

```bash
# Locate the installed file
INSTALLED=$(python3 -c "import eodag.api.product.metadata_mapping as m; print(m.__file__)")

# Apply the one-line fix
sed -i '' \
  's/band = processed_bands\[index\]$/band = processed_bands[index] if index < len(processed_bands) else {}/' \
  "$INSTALLED"

# Verify
uvx eodag search -p earth_search -c COP_DEM_GLO30_DGED \
  --box -3 41 5 44 --all --storage /tmp/dem_test
# Expected: "Returned 29 products"
```

Or as a pure-Python unit check (no network, no credentials):

```python
from eodag.api.product.metadata_mapping import normalize_bands

asset = {
    "href": "s3://copernicus-dem-30m/example.tif",
    "raster:bands": [{"data_type": "float32", "nodata": 0, "unit": "meter", "spatial_resolution": 30}],
}
normalize_bands(asset)   # IndexError before fix, succeeds after
assert "bands" in asset
assert "raster:bands" not in asset
print("OK")
```

## Reproduction

```bash
uvx eodag search -p earth_search -c COP_DEM_GLO30_DGED \
  --box -3 41 5 44 --all --storage dem_search
# Before fix: "Returned 0 products"  (IndexError silently swallowed)
# After fix:  "Returned 29 products"
```

## Changes

[...]

PR #2172, opened 2026-05-04
fix: conversion of bands in assets

jlahovnik:
avoid error if there are no eo:bands


jlahovnik:
already solved in https://github.com/CS-SI/eodag/pull/2171

PR #2173, opened 2026-05-05
fix(plugins): cop_marine asset urls

jlahovnik:
- fix for incorrect download urls for `cop_marine` items 
- also removes some superfluous code (maybe came in due to merge error)


eodag/plugins/search/cop_marine.py
sbrunato, line 1: please add a comment and also a test for that

sbrunato (changes requested):

eodag/plugins/search/cop_marine.py
jlahovnik: added

sbrunato (approved):

PR #2212, opened 2026-06-01
fix(providers): dedl pagination

cauriol:
When we do a search for an item on the dedl providers i got : {"code":"400","ticket":"d3c480e6","description":"1 error(s). page: Extra inputs are not permitted"}')

How to locally reproduce using `stac-fastapi-eodag`:
http://127.0.0.1:8000/collections/METOP_IASSND02/items?filter=federation:backends=%27dedl%27

This Pull request fix that issue

### DIFF
--- a/eodag/resources/providers.yml
+++ b/eodag/resources/providers.yml
@@ -5086,6 +5086,7 @@
         extent: '$.extent'
     pagination:
       max_limit: 100
+      next_page_token_key: token
     sort:
       sort_param_mapping:
         id: id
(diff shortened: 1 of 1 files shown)


jlahovnik (approved):

eodag/resources/providers.yml
sbrunato, line 1: pagination is still broken (cannot fetch next pages using this configuration)

sbrunato (changes requested):

eodag/resources/providers.yml
jlahovnik: The HDA returns a next page link in "links" -> entry with relation "next". I think this could probably be mapped with `next_page_url_key_path`.

cauriol: The pagination is working with next_page_token_key: token

PR #2241, opened 2026-06-17
fix(plugins): dedl required queryables

### LINKED ISSUE #2214: Queryables from dedl provider are always shown as not required
**Describe the bug**
The dedl provider has a field "required" in the queryables endpoint which indicates which queryables are required. This field is ignored by eodag and all parameters are marked with "required": "False".

**Code To Reproduce**
```py
queryables = dag.list_queryables(collection="CAMS_GAC_FORECAST", provider="dedl")
for k, v in queryables.items(): 
    print(k, v)
```

**Output**
Compete output obtained with maximal verbosity.

**Environment:**
 - Python version: `3.14`
 - EODAG version: `4.3`

**Additional context**
Add any other context about the bug here.

cauriol:
The DEDL provider exposes a required field in the queryables endpoint to indicate which queryable parameters are mandatory. However, this field was ignored by EODAG, causing all parameters to be marked with "required": false.

To fix this issue, I updated the discover_metadata method to correctly propagate the required information when calling json_field_definition_to_python. I also updated json_field_definition_to_python to store the "json_schema_required" metadata directly in field_info.metadata.

fixes https://github.com/CS-SI/eodag/issues/2214

### DIFF
--- a/eodag/plugins/search/qssearch.py
+++ b/eodag/plugins/search/qssearch.py
@@ -2268,7 +2268,9 @@ class StacSearch(PostJsonSearch):
                 # convert provider json field definition to python
                 default = kwargs.get(param, json_mtd.get("default"))
                 annotated_def = json_field_definition_to_python(
-                    json_mtd, default_value=default
+                    json_mtd,
+                    default_value=default,
+                    required=json_param in resp_as_json.get("required", []),
                 )
                 field_definition = get_args(annotated_def)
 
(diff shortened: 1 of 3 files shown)


jlahovnik (changes requested):
Can you add a test for that (or adapt an existing one)?

jlahovnik (approved):

sbrunato (approved):

PR #2247, opened 2026-06-22
fix(plugins): add datetime to queryables

cauriol:
Since EODAG v4 `datetime` is accepted in the search but it is not accepted as a filter in the queryables.

This fix translates `datetime` into a start/end parameters when appropriate. The conversion is only performed if start or end are not already present, preventing any override of existing temporal query parameters defined by the provider.

This ensures compatibility with the STAC parameters while preserving existing query behavior.

### DIFF
--- a/eodag/plugins/search/build_search_result.py
+++ b/eodag/plugins/search/build_search_result.py
@@ -65,7 +65,6 @@ from eodag.utils.dates import (
     DATE_RANGE_PATTERN,
     compute_date_range_from_params,
     format_date,
-    get_datetime,
     is_range_in_range,
     parse_date,
     parse_to_utc,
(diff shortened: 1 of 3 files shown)


jlahovnik (approved):

eodag/plugins/search/build_search_result.py
sbrunato, line 1: PR description and content are confusing. Please update to EODAG lib, without mixing with `stac-fastapi-eodag`.

`datetime` is not a queryable in EODAG lib.

sbrunato (changes requested):

eodag/plugins/search/build_search_result.py
jlahovnik: `datetime` is a queryables parameter since v4, I updated the description

sbrunato (approved):

PR #2272, opened 2026-07-15
fix: add schema_href to FederationExtension

alambare:
The Federation extension does not advertise its conformance class.

Also update the validation test to use only local schemas, avoiding schema downloads during test execution.

Validation of the Federation extension currently fails because the advertised conformance class URLs are not accessible: https://github.com/Open-EO/openeo-api/issues/598


alambare:
https://github.com/CS-SI/stac-fastapi-eodag/issues/114

sbrunato:
after discussion, we'll wait to have a valid json schema available to expose its url

PR #1822, opened 2025-09-09
perf(s3): reduce stream memory footprint and add backpressure limit

alambare:
Files were fully loaded into memory when using `AwsDownload` [stream_download()](https://eodag.readthedocs.io/en/latest/api_reference/eoproduct.html#eodag.api.product._product.EOProduct.stream_download).

The back-pressure limit prevents the explosion of the buffer size when waiting on long pending requests.

PR #2275, opened 2026-07-20
refactor: reduce memory usage in stream download from S3

jlahovnik:
During the stream_download from an S3 provider the memory usage of EODAG was strongly increasing although the data is streamed to the user and should be removed from memory afterwars. Two optimisations were identified:

- remove the reference to the futures in the `FileInfo` objects to avoid a circular reference
- introduce a maximum buffer size - if the buffer size is reached we wait for the completion of a specific future instead of the next one available

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -26,7 +26,7 @@ repos:
           - flake8-eradicate
 
 -   repo: 'https://github.com/ambv/black'
-    rev: 22.12.0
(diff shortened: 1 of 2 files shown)


jlahovnik:
PR for black update should be merged first: https://github.com/CS-SI/eodag/pull/2274

jlahovnik:
problem already addressed in https://github.com/CS-SI/eodag/pull/1822

Issue #1772, opened 2025-08-18
list unavailable providers


sbrunato:
Add a way to list unavailable / pruned providers. For the moment this information is only available through logs or a private dag attribute.

Add a `dag.unavailable_providers()` method, and document it in [Python API User Guide / Providers available](https://eodag.readthedocs.io/en/stable/notebooks/api_user_guide/2_providers_products_available.html#Providers-available)

Issue #1497, opened 2025-01-27
eodag.utils.exceptions.UnsupportedProvider: usgs is not (yet) supported for LANDSAT_C2L2


klimeto:
**Describe the bug**
I want to use EODAG to download Landsat data from USGS. After registration I have tried a simple code and the eodag lib responds with UnsupportedProvider exception. I am not usre if it is a bug or misconfiguration at my end.

**Code To Reproduce**
CLI commands or Python code snippet to reproduce the bug. Please use maximum verbosity using:

**test_eodag.py**

```py
import os
from eodag import EODataAccessGateway, setup_logging
from datetime import datetime

setup_logging(3)

config_file = "./config_usgs.yaml"
dag = EODataAccessGateway(config_file)

search_results, total_count = dag.search(
    # productType='S2_MSI_L1C',
    productType='LANDSAT_C2L2',
    geom={'lonmin': 1, 'latmin': 43.5, 'lonmax': 2, 'latmax': 44}, # accepts WKT polygons, shapely.geometry, ...
    start='2023-01-01',
    end='2023-01-15',
    provider='usgs'
)

# Search for products
products = dag.search(search_results)
print(products)

product = search_results[0]
print(product)

product_path = product.download()
print(product_path)
```

**config_usgs.yaml**

```yaml
usgs:
    priority:
    api:
        extract:
        outputs_prefix:
        dl_url_params:
        product_location_scheme:
        credentials:
            username: jeodppdw
            password: our_password
```

**Output**
```
 python test_eodag.py
2025-01-27 17:22:04,245 eodag.config                     [DEBUG   ] Loading configuration from /home/klimeto/Projects/jeodpp/eodag_test/.venv/lib/python3.10/site-packages/eodag/resources/providers.yml
2025-01-27 17:22:04,597 eodag.plugins.manager            [DEBUG   ] EcmwfApi plugin skipped, eodag[ecmwf] or eodag[all] needed
2025-01-27 17:22:04,599 eodag.plugins.manager            [DEBUG   ] UsgsApi plugin skipped, eodag[usgs] or eodag[all] needed
2025-01-27 17:22:04,720 eodag.plugins.manager            [DEBUG   ] CSWSearch plugin skipped, eodag[csw] or eodag[all] needed
2025-01-27 17:22:04,753 eodag.config                     [INFO    ] Loading user configuration from: /home/klimeto/Projects/jeodpp/eodag_test/config_usgs.yaml
[...]

PR #2341, opened 2026-09-02
feat(core): explicit error messages and doc for not-available providers

sbrunato:
Use more explicit error messages and documentation for not-available providers. All are now handled in *pruned* providers, together with the reason that caused the provider removal:
- skipped plugin because of missing dependencies
- missing credentials when auth is required for search
- missing search plugin
- missing auth plugin when auth is required for search

Raise `MisconfiguredError` when provider is missing because of lacking credentials instead of `UnsupportedProvider('provider is not (yet) supported')`:

> eodag.utils.exceptions.MisconfiguredError: provider needing auth for search was pruned because no credentials could be found

Related to https://github.com/CS-SI/eodag/issues/1497
Updates https://github.com/CS-SI/eodag/issues/1772

### DIFF
--- a/docs/getting_started_guide/configure.rst
+++ b/docs/getting_started_guide/configure.rst
@@ -20,6 +20,18 @@ for products in a provider's catalog, how to download products, etc. However, us
 to complete this configuration with additional settings, such as provider credentials. Users can
 also override pre-configured settings (e.g. the download folder).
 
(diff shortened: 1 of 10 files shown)


alambare:
@sbrunato I am skeptical on using `MisconfiguredError` on situations such as missing credentials because it is not an error but the nominal behavior. The provider is disabled because no authentication is present. I believe we should handle those situations more cleanly.

PR #2352, opened 2026-09-15
feat(product): keep STAC item links in EOProduct

### LINKED ISSUE #2334: keep STAC item links in EOProduct
Required by https://github.com/CS-SI/eodag/issues/2236

Keep STAC item links in EOProduct for `StacSearch` providers: 
- add `EOProduct.links: list[dict[str, str]]` attribute
- perform like with `assets` entry, that is temporary mapped in properties, then moved to `EOProduct.assets` attribute during `QueryStringSearch.normalize_results`
- support links in serialize / deserialize methods
- add links to `EOProduct` html representation

tunglambk:
Closes #2334.

`EOProduct` now carries the STAC item links of a search result, handled like `assets`: search plugins map them into the product properties, `QueryStringSearch.normalize_results` moves them to the new `EOProduct.links` attribute, `as_dict()` serializes them and `_import_stac_item_from_serialized()` restores them, and the HTML representation lists them.

The links are stored as they come from the item, so callers can resolve the original STAC
relations (`self`, `root`, `parent`, `alternate`, ...) on a product built from a `StacSearch` provider or from a serialized product.

`pytest tests/units/test_eoproduct.py tests/units/test_search_plugins.py` — 250 passed, and `flake8` is clean. The three new tests fail without the change (`AttributeError` on the missing attribute, and the serialized feature without the links).


eodag/plugins/search/qssearch.py
sbrunato, line 1: `properties["links"]` is empty for STAC providers. Please update `eodag/resources/stac_provider.yml` and add `links: '$.links'`  in metadata-mapping to make them available

sbrunato, line 1: in `EODataAccessGateway.serialize`, make sure that serialized collection link replaces the remote one. For the moment (with https://github.com/CS-SI/eodag/pull/2352/changes#r4013603816 applied), there are 2 collection links per item instead of 1.

sbrunato (changes requested):

sbrunato:
Thanks for the PR @tunglambk !

tunglambk:
Both review points are addressed in two new commits.

`links: '$.links'` is added to the STAC metadata mapping in `eodag/resources/stac_provider.yml`, so `product.links` is now populated for STAC providers; the regression test drives a mocked STAC response through the real mapping instead of patching `extract_properties`.

`EOProduct.as_dict` now skips provider links with `rel: collection`, so each serialized item keeps a single collection link pointing to the serialized `<collection>.json`; this is the feature written by `EODataAccessGateway.serialize`.

Ran `pytest tests/units` (785 passed) and the full collection including doctests (977 passed); the 4 remaining failures (`TestCoreInvolvingConfDir`, `test_eoproduct_download_progress_bar`) also fail on the unmodified branch and are environment-related. `flake8 --jobs=1`, `black --check`, `isort --check-only` and `pydocstyle` are clean; `mypy -p eodag` reports only 2 pre-existing errors in `config.py`/`provider.py`. Both new tests fail without the change (`links == []`; 2 collection links instead of 1).

sbrunato (approved):

